PR #4, opened 2018-03-19
Added Button component

CarloPalinckx:
In this pr:
- A storybook theme decorator to provide a theme to all stories.
- Minor tweaks to tslint config.
- Types for themed components.
- Index file for Spacer.
- Button component.


src/components/Button/Button.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { StatelessComponent } from 'react';
```
woutsluiter: Rename to `Button.template` would be nice

PR #12, opened 2018-03-23
Added Icon component

CarloPalinckx:
This PR adds:
- A bunch of icons
- Icon component which enforces proper icon usage.
- svg-inline webpack loader to allow for distributed icons.


src/components/Icon/Icon.style.tsx
```diff
@@ -0,0 +1,17 @@
...
+import styled from '../../utility/styled/styled';
+import Icon, { PropsType, SmallPropsType } from './Icon.template';
+
+const StyledIcon:StyledComponentClass<PropsType, ThemeType> = styled(Icon as ComponentType<SmallPropsType>)`
+    display: inline-block;
+    vertical-align: middle;
+    width: ${({ size }):string => size as PropsType['size'] === 'large' ? '60px' : '18px'};
+
+    svg {
+        fill: ${({ theme, color }):string => color !== undefined ? color : theme.iconFill};
```
woutsluiter: In my experience `currentColor` works really well with svg icons. For example when you have an icon in a button, it neatly inherits the color including hover/focus states.

CarloPalinckx: This is something to consider 👍. This is would mean that color is more often configured, do you think we should restrict color usage to theme properties in this case?

PR #13, opened 2018-03-26
Added BreakpointProvider

woutsluiter:


src/components/BreakpointProvider/BreakpointProvider.template.tsx
```diff
@@ -0,0 +1,70 @@
...
+type StateType = {
+    breakpoint:BreakpointType['name'];
+};
+
+type BreakpointType = {
+    size:number;
+    name:'small'|'medium'|'large';
+};
+
+class BreakpointProvider extends Component<PropsType, StateType> {
```
CarloPalinckx: The map seems a bit heavy boilerplate-wise. Maybe you could consider using an object instead? I would propose something like:
```
type BreakpointsType = {
    small:number;
    medium:number;
    large:number;
}
```
You could then also use `keyof BreakpointsType` instead of the union. This feels like it scales a bit nicer.

src/components/BreakpointProvider/BreakpointProvider.template.tsx
```diff
@@ -0,0 +1,70 @@
...
+    private sortedMap:Array<BreakpointType>;
+    private element:HTMLDivElement;
+
+    public constructor(props:PropsType) {
+        super(props);
+        this.state = { breakpoint: 'small' };
+        this.sortedMap = this.getSortedMap();
```
CarloPalinckx: There are two things to consider here:
- You are only sorting once in the constructor. This would become a problem when the props change after mounting.
- I think that this sort could be avoided by using something like reduce.

src/components/BreakpointProvider/BreakpointProvider.template.tsx
```diff
@@ -0,0 +1,70 @@
+import React, { Component } from 'react';
+
+type PropsType = {
+    map:Array<BreakpointType>;
+    render(breakpoint:BreakpointType['name']):JSX.Element;
```
CarloPalinckx: Since there is only 1 render callback you could consider using children as the render callback. This looks a bit nicer imo.

So instead of:
```
<BreakpointProvider
    render={(bp) => <div>{bp}</div>}
/>
```
It would be:
```
<BreakpointProvider>
    {(bp) => <div>{bp}</div>}
</BreakpoinrtProvider>
```

src/components/BreakpointProvider/BreakpointProvider.story.tsx
```diff
@@ -0,0 +1,41 @@
...
+];
+
+storiesOf('BreakpointProvider', module)
+    .add('default', () => (
+        <BreakpointProvider map={breakpointMap} render={(breakpoint):JSX.Element => {
+            return (
+                <StyledDiv>
+                    <div>Breakpoint map:</div>
+                    <div>
+                        <code dangerouslySetInnerHTML={{ __html: JSON.stringify(breakpointMap) }} />
```
CarloPalinckx: We shouldn't show used props like this. We should rather use a storybook addon like [knobs](https://github.com/storybooks/storybook/tree/master/addons/knobs). This provides (no pun intended) the user with a more meaningful way to tune the component props. We could add this in a different PR.

woutsluiter: Ah yes much better indeed

src/components/BreakpointProvider/BreakpointProvider.template.tsx
```diff
@@ -0,0 +1,64 @@
...
+};
+
+class BreakpointProvider extends Component<PropsType, StateType> {
+    private element:HTMLDivElement;
+
+    public constructor(props:PropsType) {
+        super(props);
+        this.state = { breakpoint: 'small' };
+    }
+
```
CarloPalinckx: Changing the props doesn't trigger a re-render yet. You might want to add a `componentWillReceiveProps` where you diff nextProps with the currentProps and call a setBreakpoint on a missmatch.

PR #17, opened 2018-04-03
Added Text Components

CarloPalinckx:
This PR adds:
- Storybook "options" add-on
- A `Heading` Component
- A `SubHeading` Component
- A `Text` Component
- Text based Theme properties


src/components/Heading/Heading.template.tsx
```diff
@@ -0,0 +1,29 @@
...
+        |3
+        |4
+        |5
+        |6;
+};
+
+const Heading:StatelessComponent<PropsType> = (props):JSX.Element => {
+    const Element = props.hierarchy !== undefined
+        ? `h${props.hierarchy}`
+        : 'span';
```
woutsluiter: A heading has a default `block` styling, so a div would seem a better choice.

CarloPalinckx: Ah right, good spot.

PR #18, opened 2018-04-05
Added FlatButton component, adjusted themeType

woutsluiter:
This pr adds:
- a FlatButton component for text style buttons.
- nested ThemeType properties for readability.

### DIFF
--- a/src/components/Button/Button.style.tsx
+++ b/src/components/Button/Button.style.tsx
@@ -1,23 +1,31 @@
 import { StyledComponentClass } from 'styled-components';
-import { ThemeType } from '../../themes';
+import { ButtonThemeType, ThemeType } from '../../themes';
 import styled from '../../utility/styled';
 import ButtonTemplate, { PropsType } from './Button.template';
 
-const applyTheme = (
-    backgroundColor:string,
-    borderRadius:string,
-    borderWidth:string,
-    borderColor:string,
-    color:string
-):string => {
+const applyTheme = (buttonTheme:ButtonThemeType):string => {
     return `
-        background-color: ${backgroundColor};
-        border-radius: ${borderRadius};
-        border: solid ${borderWidth} ${borderColor};
-        color: ${color};
+        background-color: ${buttonTheme.backgroundColor.default};
+        border-radius: ${buttonTheme.borderRadius};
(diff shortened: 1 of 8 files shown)


src/components/Button/Button.style.tsx
```diff
@@ -40,34 +48,10 @@ const Button:StyledComponentClass<PropsType, ThemeType> = styled(ButtonTemplate)
...
-                theme.buttonPrimaryBackgroundColor,
-                theme.buttonPrimaryBorderRadius,
-                theme.buttonPrimaryBorderWidth,
-                theme.buttonPrimaryBorderColor,
-                theme.buttonPrimaryColor,
-            );
+            case 'secondary': return applyTheme(theme.button.secondary);
+            case 'destructive': return applyTheme(theme.button.destructive);
+            case 'warning': return applyTheme(theme.button.warning);
+            default: return applyTheme(theme.button.primary);
```
CarloPalinckx: 💎

src/themes.ts
```diff
@@ -31,27 +18,142 @@ type ThemeType = {
...
-    buttonSecondaryBorderColor: '#d2d7e0',
-    buttonSecondaryBorderRadius: '0',
-    buttonSecondaryBorderWidth: '1px',
-    buttonSecondaryColor: '#333740',
-    buttonWarningBackgroundColor: '#FCC200',
-    buttonWarningBorderColor: '#FCC200',
-    buttonWarningBorderRadius: '0',
-    buttonWarningBorderWidth: '1px',
-    buttonWarningColor: '#fff',
+    button: {
```
CarloPalinckx: I really like the nesting, however I think we should consider what the shape of the theme will look like. In other words: 
- What would define a nesting. 
- Would we make a "common" nesting containing things like `fontFamily` and `fontSize`. 
- Is there a way (if we even should) we could enforce boundaries on those nestings? (ie: button components can only use `theme.button` etc.).

Could you maybe create an issue and write up a little proposal so we can document this?

woutsluiter: Created issue https://github.com/MyOnlineStore/bricks/issues/20

CarloPalinckx: 👍

PR #25, opened 2018-04-09
Added disabled state to Button

woutsluiter:
### This PR:

**Adds** ✨
- A `disabled` prop to `Button`

image


src/components/Button/Button.style.tsx
```diff
@@ -70,6 +70,25 @@ const Button:StyledComponentClass<PropsType, ThemeType> = styled(ButtonTemplate)
         transform: translateY(-2px);
         box-shadow: 0 9px 3px -6px rgba(51, 55, 64 ,.3);
     }
+
+    &[disabled] {
```
CarloPalinckx: Can't you just use `:disabled` here?

src/components/Button/Button.style.tsx
```diff
@@ -70,6 +70,25 @@ const Button:StyledComponentClass<PropsType, ThemeType> = styled(ButtonTemplate)
...
+        cursor: default;
+        opacity: .7;
+        transform: scale(1);
+        color: ${({ theme }):string => theme.button.secondary.color.default};
```
CarloPalinckx: I think this would be a good case where we would want to lock the styling. I think the grey works really well as a disabled color regardless of theme color. I think this would be a hard case for providing a consistent experience if we allow this to be themed since brighter colors would require a higher contrast for the "stripes".

CarloPalinckx: On second thought, we could also add it in the theme and restrict configuration there. That way we would still have a single place to store all color/font/etc. variables.

woutsluiter: I agree that a plain black and white version of a button would work well indeed, provided that there are no other greys in that theme. In our case though, we have greys in our styleguide that are slightly off and really don't match plain greys. And I would argue that you should not enforce this, as it would go against our own styleguide. However it should also not be bindedn to the secondary button. So I would prefer to have a theme object for a disabled button state. This might also be a better option for the FlatButton. Hold my beer:

```
    button: {
        ...
        primary: {
            backgroundColor: {
                default: '#5BD16A',
                hover: '#5BD16A',
                focus: '#5BD16A',
                active: '#5BD16A',
            },
            borderColor: {
                default: '#5BD16A',
                hover: '#5BD16A',
                focus: '#5BD16A',
                active: '#5BD16A',
            },
            borderRadius: '0',
            borderWidth: '1px',
            color: {
                default: '#fff',
                hover: '#fff',
                focus: '#fff',
                active: '#fff',
            },
        },
    },
    disabledButton: {
        backgroundColor1: '#EDEFF2',
        backgroundColor2: '#ECEDF0',
        borderRadius: '0',
        color: '#333740',
    },
    flatButton: {
        border`Color: {
            default: 'transparent',
            hover: '#d2d7e0',
            focus: '#d2d7e0',
            active: '#d2d7e0',
        },
        borderRadius: '0',
        borderWidth: '1px',
        color: {
            default: '#21252B',
            hover: '#21252B',
            focus: '#21252B',
            active: '#21252B',
        },`
    },
```

CarloPalinckx: This looks alright to me. Might however want to replace `disabledButton` with `buttonDisabled` for easier lookup. Same goes for `flatButton`.

src/components/Button/Button.template.tsx
```diff
@@ -6,6 +6,7 @@ type PropsType = {
     variant?:'primary' | 'secondary' | 'destructive' | 'warning';
     target?:HTMLAnchorElement['target'];
     href?:string;
+    disabled?:boolean;
```
CarloPalinckx: Since this props isn't relevant for the `FlatButton` you might want to provide a `Flatbutton.template` that guards (and provides better type discovery) against the `disabled` prop.

src/components/Button/Button.template.tsx
```diff
@@ -6,6 +6,7 @@ type PropsType = {
     variant?:'primary' | 'secondary' | 'destructive' | 'warning';
```
CarloPalinckx: This type can also extend `StyledType` with:
``` type PropsType = StyledType & {```
This adds the `className` prop too so you can drop that here too.

PR #34, opened 2018-04-20
Box compontent

woutsluiter:
### This PR:

**Adds** ✨
- `Box` component for aligning child components

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -609,23 +609,16 @@
     "@types/node": {
       "version": "9.4.7",
       "resolved": "http://registry.npmjs.org/@types/node/-/node-9.4.7.tgz",
-      "integrity": "sha512-4Ba90mWNx8ddbafuyGGwjkZMigi+AWfYLSDCpovwsE63ia8w93r3oJ8PIAQc3y8U+XHcnMOHPIzNe3o438Ywcw=="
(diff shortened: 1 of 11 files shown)


src/components/Row/Row.template.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { StatelessComponent } from 'react';
+
+type PropsType = {
```
CarloPalinckx: you can intersect `StyledType` and then you can also drop `className`.

src/components/Row/Row.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+import Row from '.';
+
+describe('Row', () => {
+    it('aligns children to the right', () => {
+        const row = shallow(<Row horizontal="right" />);
```
CarloPalinckx: You need to wrap enzyme's shallowWrapper in a `toJson` imported from `enzyme-to-jest` to get a proper snapshot.

CarloPalinckx:
So you we're talking about adding a `Column` component to this to deal with horizontal spacing right? How are you planning to handle the "flex children" ?

woutsluiter:
But now that i'm looking at it, I wonder if a separate element is necessary. Maybe a single `Flex` component would suffice with the ability to flex itself (basis, grow, shrink)

woutsluiter:
But then again, this notation seems really self explanatory 🤔

PR #36, opened 2018-04-20
Added PriceTag component

CarloPalinckx:
### This PR:

**Adds** ✨
- A `PriceTag` component.
- A few keys to the theme.

### DIFF
--- a/src/themes/MosTheme/MosTheme.theme.ts
+++ b/src/themes/MosTheme/MosTheme.theme.ts
@@ -141,6 +141,11 @@ const theme:ThemeType = {
     headingFontFamily: 'Melbourne, sans-serif',
     headingFontSize: '28px',
     iconFill: '#5bd16a',
+    priceTag: {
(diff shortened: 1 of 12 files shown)


woutsluiter:
Maybe the `sup` should also have a smaller font-size. It could be a fixed factor like `.5em`, or themable.
image

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -141,6 +141,11 @@ const theme:ThemeType = {
     headingFontFamily: 'Melbourne, sans-serif',
     headingFontSize: '28px',
     iconFill: '#5bd16a',
+    priceTag: {
+        color: '#333740',
+        fontFamily: 'Source Sans Pro, sans-serif',
+        fontSize: '28px',
```
woutsluiter: `fontWeight` also might be a useful theme option

PR #40, opened 2018-04-24
Added Nested types to theme

### LINKED ISSUE #20: Apply nesting to ThemeType
We need to define the shape/pattern for theme settings so it's clear and predicable in use.

Things to consider:
- nesting rules/pattern
- ways of enforcing boundaries on nesting

#### Proposal
https://gist.github.com/woutsluiter/d661a2348b1b2c07f2d42b80d1da8667

CarloPalinckx:
### This PR:

- [x] could use a little bit of grooming on the `ThemeType`. I feel like we could tweak this even further.

resolves #20

**Adds** ✨
- An extra export for each component that exports it's theme

**Changes** 🌀
- `MosTheme` according to: #20 
- The `ThemeType` to import every component's theme type


src/components/Button/Button.style.tsx
```diff
@@ -1,37 +1,73 @@
...
+    primary:ButtonStylesType;
+    secondary:ButtonStylesType;
+    destructive:ButtonStylesType;
+    warning:ButtonStylesType;
+    disabled:{
+        color:string;
+        backgroundColor:string;
+
+        lines:{
+            background:string;
```
woutsluiter: What about just a single `stripingColor`? You could combine that with the `backgroundColor` for the zebra striping. `lines` with only a `background` prop seems a little redundant. Plus `background` kinda deviates from `backgroundColor`

src/components/Notification/Notification.style.tsx
```diff
@@ -1,20 +1,36 @@
...
-const applyTheme = (theme:ThemeType, severity:PropsType['severity']):string => `
-    background: ${theme.notification[severity].background};
-    border-color: ${theme.notification[severity].borderColor};
-    color: ${theme.notification[severity].color};
-`;
+type VariantStyleType = {
+    background:string;
+    borderColor:string;
+    color:string;
+    fontFamily:string;
```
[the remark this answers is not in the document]
CarloPalinckx: Maybe we could consider using a `shared` key to fight duplication?

src/components/SubHeading/SubHeading.style.tsx
```diff
@@ -2,13 +2,22 @@ import { StyledComponentClass as _S } from 'styled-components';
...
+};
+
 const StyledSubHeading = styled(SubHeading)`
-    color: ${({ theme }):string => theme.subHeadingColor};
-    font-family: ${({ theme }):string => theme.subHeadingFontFamily};
-    font-size: ${({ theme }):string => theme.subHeadingFontSize};
+    color: ${({ theme }):string => theme.SubHeading.color};
+    font-family: ${({ theme }):string => theme.SubHeading.fontFamily};
+    font-size: ${({ theme }):string => theme.SubHeading.fontSize};
     font-weight: 400;
```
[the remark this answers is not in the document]
CarloPalinckx: 🐶

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -1,118 +1,144 @@
-import ThemeType from '../types/ThemeType';
+import ThemeType from '../../types/ThemeType';
+
+const sourceSans = 'Source Sans Pro, sans-serif';
+const melbourne = 'Melbourne, sans-serif';
+const green = '#5bd16a';
```
woutsluiter: For scalability I would use the color names found in http://bricks.myonlinestore.com/styleguide/colors. I don't think it will be long before we need those other variants.

src/components/Notification/Notification.style.tsx
```diff
@@ -1,20 +1,39 @@
...
-import Notification, { PropsType } from './Notification.template';
+import Notification from './Notification.template';
 
-const applyTheme = (theme:ThemeType, severity:PropsType['severity']):string => `
-    background: ${theme.notification[severity].background};
-    border-color: ${theme.notification[severity].borderColor};
-    color: ${theme.notification[severity].color};
-`;
+type VariantStyleType = {
+    background:string;
```
woutsluiter: I would use `backgroundColor` for consistency. Unless you want the option the theme notifications with `repeating-linear-gradient` or something.

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -1,172 +1,244 @@
...
+
+const fontSize = {
+    larger: '36px',
+    large: '28px',
+    medium: '22px',
+    base: '14px',
+};
+
+const green = {
+    light: '#7ce88a',
```
woutsluiter: oh I was thinking more like the other color name: `lighter1`, `darker2`, etc. so that design and frontend use the same vocabulary.

PR #42, opened 2018-04-25
Added webpack prod config

CarloPalinckx:
### This PR:

**Adds** ✨
- The `uglify-webpack-plugin`

**Changes** 🌀
- Dist artifacts to be minified and without sourcemaps
- Moves `@types/react` and `@types/react-dom` to dependecies so that they will be installed with the prod env.
- Circle configuration

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,26 +1,40 @@
-version: 2
-
-jobs:
-
-  test:
+defaults: &defaults
     docker:
       - image: circleci/node:8.9.1
     environment:
         CC_TEST_REPORTER_ID: 39ff6fab79be04ad2f3f4fe6d57d7665c761a8c0875f5f38c27624ed60ae3a22
     working_directory: ~/repo
+
+version: 2
+
+jobs:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -1,26 +1,40 @@
...
     working_directory: ~/repo
+
+version: 2
+
+jobs:
+  setup:
+    <<: *defaults
     steps:
+      - attach_workspace:
+          at: .
```
woutsluiter: Indenting seems a bit off in this file (2 and 4 spaces)

CarloPalinckx: the 4 spaces here are required by yaml :)

woutsluiter: Could be wrong here, but it looks like there are still 2 flavors used in this file. L2 has four spaces, while on L11 2 spaces are used.

CarloPalinckx: Ah I think you are right here, yaml need double spaces after a `-` property but not up there. Good spot 👍

PR #46, opened 2018-04-30
Added Box component

woutsluiter:
### This PR:

**Adds** ✨
- `Box` compontent

**Removes** 👋
- `OffsetType`


src/components/Box/Box.story.tsx
```diff
@@ -0,0 +1,66 @@
...
+`;
+
+const justifyOptions = ['flex-start', 'flex-end', 'center', 'stretch', 'space-between', 'space-around', 'space-evenly'];
+const alignItemsOptions = ['flex-start', 'flex-end', 'center', 'stretch'];
+const alignContentOptions = ['flex-start', 'flex-end', 'center', 'stretch', 'space-between', 'space-around'];
+const alignSelf = ['auto', 'flex-start', 'flex-end', 'center', 'baseline', 'stretch'];
+const directionOptions = ['row', 'row-reverse', 'column', 'column-reverse'];
+
+storiesOf('Box', module)
+    .add('Default', () => {
```
CarloPalinckx: I see that for some weird reason we gave all default cases in the stories a label of `default` would you mind changing them to `Default` in this pr?

src/components/Box/formatters/formatOffset.ts
```diff
@@ -0,0 +1,15 @@
+import { OffsetType } from '../../../types/OffsetType';
+
+const formatOffset = (offset:Array<OffsetType>):string => {
```
CarloPalinckx: I think it would be nicer if this function could just live in `Box.style.tsx`. You don't really need a specific test for this either imo. It is tied to the snapshot anyways.

src/types/OffsetType.ts
```diff
@@ -0,0 +1 @@
+export type OffsetType = 0|6|9|12|24|36|48|-6|-9|-12|-24|-36|-48|'auto';
```
CarloPalinckx: Nice, could you apply this type to Spacer as well? It should be backwards compatible.

src/components/Box/formatters/formatOffset.ts
woutsluiter: Actually I would like to suggest that we move this to utility, since `Spacer` will be using the same function. I think it would come in handy for all array to shorthand values.

src/components/Box/Box.story.tsx
```diff
@@ -0,0 +1,67 @@
...
+        const items:Array<JSX.Element> = [];
+
+        for (let i = 0; i < 6; i++) {
+            items.push(
+                <Box
+                    key={i}
+                    margin={trbl(select('Child margin', [0, 6, 12], 0, 'Child'))}
```
CarloPalinckx: If you could update `storybook-knobs` for the grouping that would be nice. You could then also remove remove the "Child" and "Parent" prefixes on the labels.

PR #63, opened 2018-05-15
Message component

woutsluiter:
### This PR:

**Adds** ✨
- `MessageStream` component
- `mountWithTheme` utility


src/components/Message/Message.story.tsx
```diff
@@ -0,0 +1,41 @@
+import { boolean } from '@storybook/addon-knobs/react';
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import Message from '.';
+
+storiesOf('Message', module)
+    .add('Success', () => (
```
CarloPalinckx: Would be nice if you could turn this into a single story that has knobs. You would have to add some tests to keep coverage though.

src/components/Message/Message.template.tsx
```diff
@@ -0,0 +1,34 @@
...
+const Message:StatelessComponent<PropsType> = (props):JSX.Element => {
+    return (
+        <div className={props.className}>
+            <Spacer
+                offset={trbl(12, 24)}
+                offsetType="inner"
+            >
+                <Text strong>{props.title}</Text>
+                <Text>{props.message}</Text>
+                <Text descriptive>{props.date}</Text>
```
CarloPalinckx: I think you forgot to add support for the action?

src/components/Message/Message.style.tsx
```diff
@@ -0,0 +1,46 @@
...
+import Message from './Message.template';
+
+type VariantStyleType = {
+    borderColor:string;
+};
+
+type MessageThemeType = {
+    common:{
+        backgroundColor:string;
+        borderColor:string;
```
CarloPalinckx: Would be preferable if the grouping behavior could be abstracted to something like a `MessageGroup` component.

src/components/MessageStream/MessageStream.style.tsx
```diff
@@ -0,0 +1,55 @@
...
+`;
+
+const MessageSeparator = styled.div`
+    border-bottom: 1px solid ${({ theme }):string => theme.MessageStream.common.borderColor};
+`;
+
+const StyledMessage = styled(Message)`
+    position: relative;
+    background: ${({ theme, read }):string => read ? theme.MessageStream.read.backgroundColor : theme.MessageStream.common.backgroundColor};
+
```
CarloPalinckx: nit: whitespace

PR #70, opened 2018-05-30
Text input

CarloPalinckx:
### This PR:

**Adds** ✨
- A `TextField` component.

**Changes** 🌀
- Moves jest mocks out of test files


src/components/TextField/TextField.style.tsx
```diff
@@ -0,0 +1,110 @@
...
+    padding: 0;
+    line-height: 1.5;
+`;
+
+type FloatingLabelProps = {
+    active: boolean;
+};
+
+const StyledFloatingLabel = styled.label`
+    transition: top 0.2s, margin-top 0.2s, font-size 0.2s, color 0.2s;
```
woutsluiter: Now we have a mix of decimal values with and without leading zero's. Don't really care which one we pick, but for the sake of consistency I think we should pick one.

src/components/TextField/TextField.style.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+const StyledWrapper = styled.div`
+    transition: border-color 0.3s;
+    border: solid 1px
+        ${({ active, theme }: WrapperProps & { theme: ThemeType }): string => {
+            return active
+                ? theme.TextField.active.common.borderColor
+                : theme.TextField.idle.common.borderColor;
+        }};
+    font-size: ${({ theme }): string => theme.TextField.idle.common.fontSize};
```
woutsluiter: This seems like a drawback of an auto formatter. Because this line is slightly shorter it's single line, but it doesn't make it 'prettier' when the rest is multiline.

src/components/TextField/TextField.style.tsx
```diff
@@ -0,0 +1,110 @@
...
+    font-family: ${({ theme }): string =>
+        theme.TextField.idle.common.fontFamily};
+    border-radius: ${({ theme }): string =>
+        theme.TextField.idle.common.borderRadius};
+    display: inline-block;
+    position: relative;
+    padding: 0.3em 12px 0.4em 12px;
+    cursor: text;
+    ${({ feedback, theme }): string =>
```
woutsluiter: This is also inconsistent with the function in `StyledFloatingLabel` in the same file

src/components/TextField/TextField.style.tsx
woutsluiter: Transitioning `font-size` and `margin` makes it kind of a jerky animation. In the monolith it's transitioned on transform which makes it a lot smoother.

src/components/TextField/TextField.style.tsx
```diff
@@ -0,0 +1,110 @@
...
+        message: string;
+    };
+};
+
+const StyledWrapper = styled.div`
+    transition: border-color 0.3s;
+    border: solid 1px
+        ${({ active, theme }: WrapperProps & { theme: ThemeType }): string => {
+            return active
+                ? theme.TextField.active.common.borderColor
```
woutsluiter: A focus state with just a darker border-color might not be sufficient. Just checked with Johan and a focus state also has an outer box-shadow.

src/components/TextField/TextField.story.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public componentDidMount(): void {
+        button('trigger focus', () => {
+            this.ref.focus();
+        });
+    }
+
+    public render(): JSX.Element {
+        return this.props.withFeedback ? (
```
nicokoelewijn: If around the whole component causes a lot of redundant code, wouldn't an if around the feedback be better?

src/components/TextField/TextField.style.tsx
```diff
@@ -0,0 +1,115 @@
...
+};
+
+const StyledInput = styled.input`
+    position: relative;
+    width: 100%;
+    border: none;
+    background: transparent;
+    font-size: inherit;
+    display: block;
+    margin: 1em 0 0 0;
```
nicokoelewijn: just a `margin-top: 1em;` would suffice here

PR #74, opened 2018-06-04
Added offset prop to Popover

woutsluiter:
### This PR:

**Adds** ✨
- `offset` prop to `Popover` for a custom offset
- `flex-grow` to the PopoverAnchor styling


src/components/Popover/Popover.template.tsx
```diff
@@ -58,7 +64,7 @@ const Popover: StatelessComponent<PropsType> = (props): JSX.Element => {
                         }
                         modifiers={{
                             offset: {
-                                offset: '0 16px',
+                                offset: props.offset !== undefined ? `${props.offset}px` : '16px',
```
CarloPalinckx: the offset value of `0 16px` seems to make the arrow spawn from the middle of the anchor. Somehow `0 16px` seems to restore previous behavior.

src/components/Popover/Popover.template.tsx
```diff
@@ -41,7 +47,7 @@ const Popover: StatelessComponent<PropsType> = (props): JSX.Element => {
             <Manager>
                 <Reference>
                     {({ ref }: ReferenceChildrenProps): JSX.Element => (
-                        <PopoverAnchor innerRef={ref}>
+                        <PopoverAnchor stretch={props.stretch !== undefined ? props.stretch : false} innerRef={ref}>
```
CarloPalinckx: This can be multine. Would you mind turning on prettier on save?

src/components/Popover/Popover.style.tsx
```diff
@@ -2,10 +2,21 @@ import { ClassAttributes as _C, HTMLAttributes as _H } from 'react';
 import { StyledComponentClass as _S } from 'styled-components';
 import _T from '../../types/ThemeType';
 import styled from '../../utility/styled';
-import { PopoverArrowPropsType } from './Popover.template';
+import { PopoverAnchorPropsType, PopoverArrowPropsType } from './Popover.template';
 
 const PopoverAnchor = styled.div`
-    display: inline-block;
+    ${({ stretch }: PopoverAnchorPropsType): string => {
+        if (stretch === true) {
```
CarloPalinckx: Can't this just be solved by wrapping the anchor in a `<Box/>` that does the stretching?

woutsluiter: I'll check it out now in the NotificationCenter story

src/components/Popover/Popover.template.tsx
woutsluiter: will do

src/components/Popover/Popover.template.tsx
woutsluiter: Looks like I forgot the zero. Should be `0 ${props.offset}px` : '0 16px'

src/components/Popover/Popover.style.tsx
woutsluiter: Alas, it won't go full width

PR #75, opened 2018-06-05
Renaming

### LINKED ISSUE #67: Rethink file naming
Maybe we could lose the Component name prefixes.
I really like the way spectrum does [this](https://github.com/withspectrum/spectrum/tree/alpha/src/components/buttons). An example (FoldOut component) would be:

```
- FoldOut
    - index.tsx
    - story.tsx
    - style.tsx
    - test.tsx
```

In this scenario the "template" file would always be the exporting file as well which might help us with the decision making on when/how to cross import styled component/components.

nicokoelewijn:
### This PR:

- [x] wait for #70 to be merged.

resolves #67 

**Adds** ✨
- Edited few test cases for button and messagestream

**Changes** 🌀
- Template.tsx is now used as index

**Removes** 👋
- Component name in filenames
- Index.ts


.codeclimate.yml
```diff
@@ -39,8 +39,12 @@ exclude_patterns:
     - '**/*_test.go'
```
CarloPalinckx: You are getting a few flags from CodeClimate. I think we can loosen the configuration a bit.

You can:
✔️  bump the allowed complexity up to 8
✔️  increase allow function length to 80
✔️  increase allowed code duplication to 3

nicokoelewijn: 👍

PR #76, opened 2018-06-07
Added ButtonChain

CarloPalinckx:
### This PR:

**Adds** ✨
- A `ButtonChain` component that allows you to require a confirmation of an action.


CarloPalinckx:
According to UX this is not really a pattern we want to support. Closing this for now.

PR #78, opened 2018-06-08
Radiobutton

### LINKED ISSUE #31: Component Wishlist ✨🙏
- [x] **Flex/Row**
Provides an abstraction for the position of components.
- [x] **ButtonGroup**
Could help with alignment of multiple buttons for responsive views.

- [x] **Scrollable**
Would be nice if we could provide a consistent way to add scrolling behavior to components.

- [x] **Modal**
For opening/closing pop-ups.

- [x] **Toaster**
For showing fixed notifications.

- [x] **Form elements**
We could use a set of form elements.

- [x] **Separator**
A component that eithers wraps and provides a border or that is rendered without children to provide just a seperating line.

nicokoelewijn:
### This PR:

**Adds** ✨
- RadioButton
- RadioButtonGroup

### DIFF
--- a/src/components/RadioButton/index.tsx
+++ b/src/components/RadioButton/index.tsx
@@ -1,0 +1,44 @@
+import React, { StatelessComponent } from 'react';
+import trbl from '../../utility/trbl';
+import Spacer from '../Spacer';
+import Text from '../Text';
+import StyledRadioButton, { StyledRadioButtonSkin, StyledRadioWrapper } from './style';
+
+type PropsType = {
+    checked: boolean;
+    value: string;
+    name: string;
+    label: string;
+    changeHandler(change: { checked: boolean; value: string }): void;
+};
+
+const RadioButton: StatelessComponent<PropsType> = (props): JSX.Element => {
+    const handleChange = (): void => {
+        props.changeHandler({
+            checked: !props.checked,
+            value: props.value,
+        });
+    };
+
+    return (
+        <StyledRadioWrapper onClick={handleChange}>
+            <Spacer offset={trbl(0, 9, 0, 0)}>
+                <StyledRadioButtonSkin checked={props.checked}>
+                    <StyledRadioButton
+                        onChange={handleChange}
+                        checked={props.checked}
+                        type="radio"
+                        name={props.name}
+                        value={props.value}
+                    />
+                </StyledRadioButtonSkin>
+            </Spacer>
+            <Text>
+                <label htmlFor={props.name}>{props.label}</label>
+            </Text>
+        </StyledRadioWrapper>
+    );
+};
+
+export default RadioButton;
+export { PropsType };
--- a/src/components/RadioButton/story.tsx
+++ b/src/components/RadioButton/story.tsx
@@ -1,0 +1,16 @@
+import { boolean } from '@storybook/addon-knobs/react';
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import RadioButton from '../RadioButton';
+
+storiesOf('RadioButton', module).add('Default', () => {
+    return (
+        <RadioButton
+            checked={boolean('checked', true)}
+            name="demo"
+            label="foo"
+            value="demo2"
+            changeHandler={(): void => undefined}
+        />
+    );
+});
--- a/src/components/RadioButton/style.tsx
+++ b/src/components/RadioButton/style.tsx
@@ -1,0 +1,58 @@
+import _R from 'react';
+import { StyledComponentClass as _S } from 'styled-components';
+import _T from '../../types/ThemeType';
+import styled, { withProps } from '../../utility/styled';
+
+type RadioButtonPropsType = {
+    checked: boolean;
+};
+
+type RadioButtonThemeType = {
+    idle: {
+        boxShadow: string;
+        borderColor: string;
+        backgroundColor: string;
+    };
+    active: {
+        boxShadow: string;
+        borderColor: string;
+        backgroundColor: string;
+    };
+    hover: {
+        boxShadow: string;
+    };
+};
+
+const StyledRadioWrapper = styled.div`
+    display: flex;
+`;
+
+const StyledRadioButton = withProps<RadioButtonPropsType, HTMLInputElement>(styled.input)`
+    position: relative;
+    display: none;
+`;
+
+const StyledRadioButtonSkin = withProps<RadioButtonPropsType, HTMLDivElement>(styled.div)`
+    width: 17px;
+    height: 17px;
+    border-radius: 100%;
+    transition: box-shadow 300ms ease, border 300ms ease;
+    background-color: ${({ theme }): string => theme.RadioButton.idle.backgroundColor};
+
+    border: 1px solid ${({ theme, checked }): string =>
+        checked ? theme.RadioButton.active.borderColor : theme.RadioButton.idle.borderColor};
+
+    box-shadow: ${({ theme, checked }): string =>
+        checked ? theme.RadioButton.active.boxShadow : theme.RadioButton.idle.boxShadow};
+
+    &:hover{
+        box-shadow: ${({ theme }): string => `inset 0 0 0 0 transparent, ${theme.RadioButton.hover.boxShadow}`};
+        box-shadow: ${({ theme, checked }): string =>
+            checked
+                ? `${theme.RadioButton.active.boxShadow}, ${theme.RadioButton.hover.boxShadow}`
+                : `${theme.RadioButton.idle.boxShadow}, ${theme.RadioButton.hover.boxShadow}`};
+    };
+`;
+
+export default StyledRadioButton;
+export { StyledRadioButton, StyledRadioButtonSkin, StyledRadioWrapper, RadioButtonThemeType };
--- a/src/components/RadioButton/test.tsx
+++ b/src/components/RadioButton/test.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import RadioButton from '.';
+import { mountWithTheme } from '../../utility/styled';
+import StyledRadioButton from './style';
+
+describe('RadioButton', () => {
+    it('should not have box-shadow when not checked', () => {
+        const radioButton = mountWithTheme(
+            <RadioButton name="demo" label="foo" checked={false} value="bar" changeHandler={(): void => undefined} />,
+        );
+
+        /* tslint:disable */
+        (expect(radioButton.find(StyledRadioButton)) as any).not.toHaveStyleRule('box-shadow', '#fff');
+        /* tslint:enable */
+    });
+
+    it('should be able to change checked value', () => {
+        const mockHandler = jest.fn();
+        const radioButton = mountWithTheme(
+            <RadioButton name="demo" label="foo" checked={false} value="bar" changeHandler={mockHandler} />,
+        );
+
+        radioButton.find(StyledRadioButton).simulate('change');
+
+        expect(mockHandler).toHaveBeenCalledWith({
+            checked: true,
+            value: 'bar',
+        });
+    });
+});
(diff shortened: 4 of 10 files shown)


src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -291,6 +295,12 @@ const theme:ThemeType = {
             fontWeight: '700',
         },
     },
+    RadioButton: {
+        common: {
```
CarloPalinckx: Would be nice if there was a way to distinguish styling for active and idle states. You now have the idle border hardcoded at `rgba(0,0,0,0.25)` would be nice if you could configure this in the theme as well. Imagine a site with a black background for instance, there you might want to provide a white border for the idle state. This would mean that you could move `border` out of common and turn the theme into something like:

```
RadioButton: {
    common: {
        boxShadow:string;
    }
    idle: {
        border:string;
    }
    checked: {
        border:string;
    }
}
```

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+type RadioButtonThemeType = {
+    common: {
+        boxShadow: string;
+        border: string;
+    };
+}
+
+const StyledRadioWrapper = styled.div`
+display: flex;
```
CarloPalinckx: Some whitespace here would be nice :)

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,41 @@
...
+`
+
+const StyledRadioButton = withProps<RadioButtonPropsType, HTMLInputElement>(styled.input) `
+position: relative;
```
CarloPalinckx: idem

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,41 @@
...
+display: none;
+`
+
+const StyledRadioButtonSkin = withProps<RadioButtonPropsType, HTMLDivElement>(styled.div) `
+background-color: #FFF;
```
CarloPalinckx: idem

src/components/RadioButton/test.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import RadioButton from '.';
+import { mountWithTheme } from '../../utility/styled';
+import StyledRadioButton from './style';
+
+describe('RadioButton', () => {
+
```
CarloPalinckx: extranous whitespace

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,41 @@
...
+width: 17px;
+height: 17px;
+border: 1px solid rgba(0, 0, 0, .25);
+border-radius: 100%;
+transition: box-shadow 0.25s ease;
+${({ theme, checked }): string => checked
```
CarloPalinckx: Whitespacing is a bit off here as well

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,41 @@
...
+        ? `box-shadow: inset 0px 0px 0px 5.5px ${theme.RadioButton.common.boxShadow};
+           transition: box-shadow 0.25s ease;`
+        : ''
+    }
+`
```
CarloPalinckx: Would be nice if the radiobutton could have some hover styling. The configured border color also doesn't seem to be applied when active.

src/components/RadioButton/RadioButton.story.tsx
```diff
@@ -0,0 +1,27 @@
...
+import React from 'react';
+import RadioButton from '../RadioButton';
+
+storiesOf('RadioButton', module)
+    .add('default', () => {
+        return (
+            <RadioButton
+                name="demo"
+                label="foo"
+                checked={false}
```
CarloPalinckx: I think this would be a cleaner form of documentation if you could keep it at one story where you can toggle the checked props with a knob. Check https://github.com/storybooks/storybook/tree/master/addons/knobs for more info on knobs.

CarloPalinckx: You might need to add a small case to your tests to cover the unchecked styling tho.

src/components/RadioButton/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+                    <StyledRadioButton
+                        onChange={handleChange}
+                        checked={props.checked}
+                        type="radio"
+                        name={props.name}
+                        value={props.value}
+                    />
+                </StyledRadioButtonSkin>
+            </Spacer>
+            <label htmlFor={props.name}>
```
CarloPalinckx: you could wrap this label in a `Text` component to provide it with all the needed styling. Or you could make the label styling configurable in the theme. Your call.

src/components/RadioButton/style.tsx
CarloPalinckx: I also think this position relative doesn't really do anything because it has a display none anyways 🤷‍♂️

src/components/RadioButtonGroup/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+};
+
+const RadioButtonGroup: StatelessComponent<PropsType> = (
+    props,
+): JSX.Element => {
+    return (
+        <BreakpointProvider
+            breakpoints={{
+                small: 0,
+                medium: 0,
```
CarloPalinckx: you can change this to 375. You are relying on the defaulting behavior now because the `medium` and `small` breakpoints are equal here. You can turn the large into anything above 375. Maybe something like 800.

src/components/RadioButtonGroup/index.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { Children, StatelessComponent } from 'react';
+import trbl from '../../utility/trbl';
+import Box from '../Box';
+import BreakpointProvider from '../BreakpointProvider';
+
+type PropsType = {
+    selectedValue?: boolean;
```
CarloPalinckx: You can remove the selected property since it is unused.

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,78 @@
...
+    styled.input,
+)`
+    position: relative;
+    display: none;
+`;
+
+const StyledRadioButtonSkin = withProps<RadioButtonPropsType, HTMLDivElement>(
+    styled.div,
+)`
+    ${({ theme, checked }): string =>
```
CarloPalinckx: Instead of one big interpolated function you might want to split this into  function per css rule, not only does this keep the string clean. It also helps with performance. Now, everytime checked changes the entire string block has to be evaluated instead of just a since function.

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -50,7 +50,7 @@ const yellow = {
     darker2: '#edb107',
 };
```
CarloPalinckx: nit: all the values without a leading zero (ie: `.2`) should have a leading zero (ie: `0.2`). Prettier doesn't always find these when they are inside a string.

src/components/RadioButton/RadioButton.story.tsx
```diff
@@ -0,0 +1,16 @@
+import { boolean } from '@storybook/addon-knobs/react';
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import RadioButton from '../RadioButton';
+
+storiesOf('RadioButton', module).add('is defaultChecked', () => {
```
CarloPalinckx: Name of the story can just be `Default` here

src/components/RadioButtonGroup/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+            {(breakpoint): JSX.Element => {
+                const direction = breakpoint === 'small' ? 'column' : 'row';
+
+                return (
+                    <Box
+                        direction={direction}
+                        justifyContent="flex-start"
+                        alignItems="stretch"
+                        wrap
+                        margin={trbl(6)}
```
CarloPalinckx: would be nice is you could only handle the spacing in between for now. This is the same approach as we did with the ButtonGroup, since the space around could be variable to the place it will be used. Which has worked out pretty good so far. That would mean give this `trbl(0,-6)` and the children `trbl(0, 6)`

src/components/RadioButton/story.tsx
```diff
@@ -0,0 +1,16 @@
+import { boolean } from '@storybook/addon-knobs/react';
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import RadioButton from '../RadioButton';
+
+storiesOf('Default', module).add('is defaultChecked', () => {
```
CarloPalinckx: Default -> RadioButton, is defaultChecked -> Default

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,68 @@
...
+const StyledRadioButton = withProps<RadioButtonPropsType, HTMLInputElement>(styled.input)`
+    position: relative;
+    display: none;
+`;
+
+const StyledRadioButtonSkin = withProps<RadioButtonPropsType, HTMLDivElement>(styled.div)`
+    width: 17px;
+    height: 17px;
+    border-radius: 100%;
+    transition: box-shadow 0.25s ease;
```
CarloPalinckx: We prefer units of ms here, The default animation speed is 300ms as well.

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,68 @@
...
+        box-shadow: ${({ theme }): string => theme.RadioButton.hover.boxShadow};
+    }
+
+    ${({ theme, checked }): string => `
+        ${
+            checked
+                ? `
+                    &:hover{
+                        box-shadow: ${theme.RadioButton.activeHover.boxShadow};
+                        transition: box-shadow 0.25s ease;
```
CarloPalinckx: You can remove this transition

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,68 @@
...
+                    }
+                    box-shadow: ${theme.RadioButton.active.boxShadow};
+                    border: 1px solid ${theme.RadioButton.active.borderColor};
+                    transition: box-shadow 0.25s ease;
```
CarloPalinckx: You can remove this transition

src/components/RadioButton/style.tsx
CarloPalinckx: You can also add border to the transition list here.

nicokoelewijn:
👍

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,66 @@
...
+        boxShadow: string;
+        borderColor: string;
+        backgroundColor: string;
+    };
+    active: {
+        boxShadow: string;
+        borderColor: string;
+        backgroundColor: string;
+    };
+    activeHover: {
```
CarloPalinckx: You can drop this part of the theme since you can get this data from active and hover states combined.

src/components/RadioButton/style.tsx
```diff
@@ -0,0 +1,66 @@
...
+    height: 17px;
+    border-radius: 100%;
+    transition: box-shadow 300ms ease, border 300ms ease;
+    background-color: ${({ theme }): string => theme.RadioButton.idle.backgroundColor};
+    border: 1px solid ${({ theme }): string => theme.RadioButton.idle.borderColor};
+    &:hover{
+        box-shadow: ${({ theme }): string => theme.RadioButton.hover.boxShadow};
+    }
+
+    ${({ theme, checked }): string => `
```
CarloPalinckx: Somehow the hover shadow transition isn't working when the radio is checked. Can't really pinpoint the origin tbh.

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -314,6 +314,21 @@ const theme: ThemeType = {
...
+        idle: {
+            backgroundColor: silver.base,
+            boxShadow: green.base,
+            borderColor: silver.darker4,
+        },
+        hover: {
+            boxShadow: '0 0 0 4px rgba(107, 222, 120, 0.4)',
+        },
+        active: {
+            boxShadow: 'inset 0px 0px 0px 5.5px #6bde78',
```
CarloPalinckx: you can replace `#6bde78` with `green.base`

PR #83, opened 2018-06-12
Added Badge

woutsluiter:
### This PR:

**Adds** ✨
- `Badge`

### DIFF
--- a/src/components/Badge/index.tsx
+++ b/src/components/Badge/index.tsx
@@ -1,0 +1,8 @@
+import React, { SFC } from 'react';
+import StyledBadge from './style';
+
+const Badge: SFC<{}> = (props): JSX.Element => {
+    return <StyledBadge>{props.children}</StyledBadge>;
+};
+
+export default Badge;
--- a/src/components/Badge/story.tsx
+++ b/src/components/Badge/story.tsx
@@ -1,0 +1,6 @@
+import { text } from '@storybook/addon-knobs/react';
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import Badge from '.';
+
+storiesOf('Badge', module).add('Default', () => <Badge>{text('text', '1')}</Badge>);
--- a/src/components/Badge/style.tsx
+++ b/src/components/Badge/style.tsx
@@ -1,0 +1,27 @@
+import _R from 'react';
+import { StyledComponentClass as _S } from 'styled-components';
+import _T from '../../types/ThemeType';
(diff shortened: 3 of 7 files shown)


src/components/Badge/index.tsx
```diff
@@ -0,0 +1,8 @@
+import React, { SFC } from 'react';
+import StyledBadge from './style';
+
+const Badge: SFC = (props): JSX.Element => {
+    return <StyledBadge>{props.children}</StyledBadge>;
+};
+
+export default Badge;
```
nicokoelewijn: index.tsx can be removed and style.tsx renamed to index.tsx

src/components/Badge/story.tsx
```diff
@@ -0,0 +1,6 @@
+import { text } from '@storybook/addon-knobs/react';
+import { storiesOf } from '@storybook/react';
+import React from 'react';
+import Badge from '.';
+
```
nicokoelewijn: There's no story about how this component is used in combination with an icon or text behind it.

src/components/Badge/style.tsx
```diff
@@ -0,0 +1,27 @@
...
+    min-height: 18px;
+    padding: 4px 6px;
+    border-radius: 9px;
+    background: ${({ theme }): string => theme.Badge.backgroundColor};
+    font-family: ${({ theme }): string => theme.Text.default.fontFamily};
+    font-size: 10px;
+    line-height: 1;
+    color: ${({ theme }): string => theme.Badge.color};
+    white-space: nowrap;
+`;
```
nicokoelewijn: the badge should either support a link/action or be clickthrough `pointer-events: none;`

src/components/Badge/story.tsx
woutsluiter: There is a textfield knob where you can edit the content, so I don't really see the need for it. And I don't think this element is suitable for Icons, or at least not the size icons we have now. You'd need an even smaller size like 10px or something.

CarloPalinckx: @woutsluiter  👀 _behind it._ 👀

woutsluiter: Oh geez need more coffee

woutsluiter: Ok so since there is no real usage pattern (yet) I'll keep it inline-block so it just aligns on the baseline by default. If a more obvious pattern emerges we can always implement it.

PR #86, opened 2018-06-18
Switch

nicokoelewijn:
### This PR:

**Adds** ✨
Toggle component


src/components/Toggle/style.tsx
```diff
@@ -0,0 +1,80 @@
...
+};
+
+const StyledToggle = withProps<StyledToggleType, HTMLInputElement>(styled.input)`
+    opacity: 0;
+    height: 0;
+    width: 0;
+`;
+
+const StyledToggleSkin = withProps<StyledToggleSkinType, HTMLDivElement>(styled.div)`
+    width: 35px;
```
woutsluiter: Just checked the sizes with Johan. The skin should be 48px in width and 18px in height (including borders). The toggle itself 24 x 24 (including borders).

src/components/Toggle/style.tsx
```diff
@@ -0,0 +1,80 @@
...
+        transition: all 100ms;
+        content: "";
+        width: 19px;
+        height: 19px;
+        top: -3px;
+        position: absolute;
+        border-radius: 100%;
+        background-color: ${({ theme }): string => theme.Toggle.idle.backgroundColor};
+        border: 1px solid ${({ theme }): string => theme.Toggle.idle.borderColor};
+        transform: ${({ checked }): string => (checked ? 'translateX(11px)' : 'translateX(-10px)')};
```
woutsluiter: In the design the toggle aligns with the edge of the ToggleSkin

src/components/Toggle/style.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    &::after {
+        transition: all 100ms;
+        content: "";
+        width: 19px;
+        height: 19px;
+        top: -3px;
+        position: absolute;
+        border-radius: 100%;
+        background-color: ${({ theme }): string => theme.Toggle.idle.backgroundColor};
```
woutsluiter: In the design the toggle is different from the the skin: `silver.lighter1`

src/components/Toggle/style.tsx
```diff
@@ -0,0 +1,80 @@
...
+            : `
+            background-color: ${theme.Toggle.idle.backgroundColor};
+            border: 1px solid ${theme.Toggle.idle.borderColor}
+        `}
+    `;
+
+const StyledToggleWrapper = styled.div`
+    display: flex;
+
+    &:hover ${StyledToggleSkin} {
```
woutsluiter: Also checked with Johan. In contrast to what he said before, the focus should not be the same as the hover. Other input elements don't have a hover state (yet).

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -415,6 +415,19 @@ const theme: ThemeType = {
             borderColor: grey.lighter1,
         },
     },
+    Toggle: {
+        idle: {
+            borderColor: silver.darker4,
+            backgroundColor: silver.base,
+        },
+        hover: {
+            boxShadow: '0 0 0 4px rgba(107,222,120,0.4)',
```
woutsluiter: CS: space after comma

nicokoelewijn: The test uses the theme for comparison:
```
(expect(toggle.find(StyledToggleSkin)) as any).toHaveStyleRule(
  'background', 
  mosTheme.Toggle.focus.boxShadow
);
```
Unfortunately the .babelrc is not loaded: https://github.com/kulshekhar/ts-jest/issues/480
Therefore the css from the component is minified. As workaround the css in the theme is also minified (`boxShadow: '0 0 0 4px rgba(107,222,120,0.4)'`) so the comparison doesn't break.

src/components/Toggle/style.tsx
```diff
@@ -0,0 +1,105 @@
...
+    ${({ theme, elementFocus }): string => (elementFocus ? `box-shadow: ${theme.Toggle.focus.boxShadow};` : '')}
+
+    background: ${({ theme, checked, disabled }): string => {
+        if (disabled && checked) {
+            return theme.Toggle.checkedDisabled.background;
+        }
+        if (!disabled && checked) {
+            return theme.Toggle.checked.background;
+        }
+
```
CarloPalinckx: This misses a `!disabled && !checked` case.

src/components/Toggle/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+                            value={this.props.value}
+                            checked={this.props.checked}
+                            disabled={this.props.disabled ? this.props.disabled : false}
+                            error={this.props.error ? this.props.error : false}
+                            type="checkbox"
+                            onChange={this.handleChange}
+                        />
+                    </StyledToggleSkin>
+                </Spacer>
+                <Text>
```
CarloPalinckx: You can remove the label from this component. We can't really guarantee that the label can be used like that so you can drop this for now. If you expose `id` as a prop we could use it as the target for `htmlFor` later on when there is a seperate `Label` component.

PR #88, opened 2018-06-21
Popoverfixes

woutsluiter:
### This PR:

**Changes** 🌀
- `Popover` positioning

### DIFF
--- a/src/components/Popover/index.tsx
+++ b/src/components/Popover/index.tsx
@@ -10,6 +10,7 @@ type PropsType = {
     isOpen: boolean;
     fixed?: boolean;
     offset?: number;
+    stretch?: boolean;
     renderContent(): JSX.Element | string;
 };
 
@@ -19,7 +20,9 @@ const Popover: StatelessComponent<PropsType> = (props): JSX.Element => {
             <Manager>
                 <Reference>
                     {({ ref }: ReferenceChildrenProps): JSX.Element => (
-                        <PopoverAnchor innerRef={ref}>{props.children}</PopoverAnchor>
+                        <PopoverAnchor innerRef={ref} stretch={props.stretch}>
+                            {props.children}
+                        </PopoverAnchor>
                     )}
                 </Reference>
                 <TransitionAnimation show={props.isOpen} animation="fade">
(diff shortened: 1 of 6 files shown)


src/components/Popover/story.tsx
```diff
@@ -88,6 +90,7 @@ storiesOf('Popover', module).add('Default', () => (
             ) as PlacementType
         }
         fixed={boolean('fixed', false)}
+        offset={number('offset', 16)}
```
nicokoelewijn: changing the value doesn't update the preview and reloads page on submit. Not shure if this is a bug in Knobs, but would be nice if this can be fixed.

src/components/Popover/index.tsx
```diff
@@ -19,7 +20,9 @@ const Popover: StatelessComponent<PropsType> = (props): JSX.Element => {
             <Manager>
                 <Reference>
                     {({ ref }: ReferenceChildrenProps): JSX.Element => (
-                        <PopoverAnchor innerRef={ref}>{props.children}</PopoverAnchor>
+                        <PopoverAnchor innerRef={ref} stretch={props.stretch}>
+                            {props.children}
```
nicokoelewijn: To much content on a small screen can cause not viewable content. Perhaps the scrollbox could solve this problem.

src/components/Popover/story.tsx
woutsluiter: You have to trigger the Popover again to se the effect. It only calculates the position once.

woutsluiter: The submit is just something the browser does when you use the enter key in a form. In the case of Knobs it only seems to happen in the number fields.

src/components/Popover/index.tsx
woutsluiter: This is indeed a job for Scrollbox. This combination is already in use in the mono repo.

src/components/Popover/index.tsx
```diff
@@ -10,6 +10,7 @@ type PropsType = {
     isOpen: boolean;
     fixed?: boolean;
     offset?: number;
+    stretch?: boolean;
```
CarloPalinckx: Could you add this prop to the knobs?

PR #97, opened 2018-07-04
Scale variations

### LINKED ISSUE #95: Text variations
The `Text`, `Heading`, `Subheading` components could use more variation to match the typographic scale. The following could be added:

### Text

- small
- base
- large

The `Text` could also use a way to vertically/horizontally align its children as well.

### Heading

The Subheading and Heading components can be merged. We could use the hierarchy prop to determine the styling of the heading and provide an alternative way of rendering custom elements(for seo purposes). A range of 1-6 should be enough for now.

robinvdv12:
### This PR:

resolves #95

**Adds** ✨
- `small`, `base` and `large` variations on Text Component.
- `element` prop on Heading Component which determine if the component will get rendered as a `h1`, `h2`, etc, `div`, `span` or `p` element.

**Changes** 🌀
- All the hierarchies of the Heading Component are now themable. (According to those [Headings](http://bricks.myonlinestore.com/styleguide/typography)).

**Removes** 👋
- SubHeading Component.


CarloPalinckx:
Overall, very nice PR! 🎉  I think we could use this pr to improve the usage of the Heading a bit. Your current changes now have a `subHeading` prop on the `Heading`. I think it would nice if drop the term `subheading` completely because it only makes a difference in styling. Maybe we could just make all the available hierarchies themable? (1-6). That way we could make the split in heading/subheading by just providing other values in the theme. Does that make sense?

src/components/ScrollBox/story.DontTest.tsx
```diff
@@ -93,7 +92,7 @@ class Demo extends Component<{}, DemoStateType> {
                     </ScrollBox>
                 </StyledDiv>
                 <Spacer offset={trbl(48, 0, 12)}>
-                    <SubHeading>Scroll event</SubHeading>
+                    {/* <SubHeading>Scroll event</SubHeading> */}
```
CarloPalinckx: You could replace this with a `Heading`

robinvdv12:
@CarloPalinckx Yes, makes sense. I will start working on it tomorrow!

src/components/Heading/index.tsx
```diff
@@ -3,11 +3,65 @@ import { StyledComponentClass as _S } from 'styled-components';
...
+type HeadingThemeType = StyledType & {
+    default: {
+        color: string;
+        fontFamily: string;
+        fontSize: string;
+        fontWeight: string;
+        lineHeight: string;
+        textTransform: string;
+    };
+    hierarchy: [
```
CarloPalinckx: instead of an array you could turn this into an object, maybe something like: 
```
{
h1: {}
h2: {}
h3: {}
h4: {}
...etc
}
```

src/components/Heading/story.tsx
```diff
@@ -5,8 +5,12 @@ import Heading, { PropsType } from '.';
 
 storiesOf('Heading', module)
     .add('Default', () => <Heading>This is a div element.</Heading>)
```
CarloPalinckx: You can also remove this story in favor of the Hierarchy one, it could be the default as well I guess.

src/components/Heading/index.tsx
```diff
@@ -3,11 +3,65 @@ import { StyledComponentClass as _S } from 'styled-components';
...
+            fontFamily: string;
+            fontSize: string;
+            fontWeight: string;
+            lineHeight: string;
+            textTransform: string;
+        }
+    ];
 };
 
 type PropsType = StyledType & {
```
CarloPalinckx: We still need a way to render a `div` as well as the styling of a hierarchy. ATM you can only render a div when you skip the hierarchy prop which results in the default styling.

CarloPalinckx: Maybe an extra prop that provides an opt-in for an element.

```
<Heading hierachy="1" element="div">Foo</Heading>
```

something like this maybe?

robinvdv12: Sow the `hierarchy` prop is just for styling? And the `element` prop expects an param (div, h1, h2, etc.) which decide if the heading will be a div, h1, h2, etc.?

CarloPalinckx: Imo, the default behavior should be:
```
hierarchy=1 -> h1 with styling for hierarchy 1
hierarchy=2 -> h2 with styling for hierarchy 2
hierarchy=3 -> h3 with styling for hierarchy 3
hierarchy=4 -> h4 with styling for hierarchy 4
hierarchy=5 -> h5 with styling for hierarchy 5
hierarchy=6 -> h6 with styling for hierarchy 6
```
and with a provided element:
```
hierarchy=1 element="div" -> div with styling for hierarchy 1
hierarchy=2 element="div" -> div with styling for hierarchy 2
hierarchy=3 element="div" -> div with styling for hierarchy 3
hierarchy=4 element="div" -> div with styling for hierarchy 4
hierarchy=5 element="div" -> div with styling for hierarchy 5
hierarchy=6 element="div" -> div with styling for hierarchy 6
```

This way we decouple the "seo" value from the styling.

Instead of my previous suggestion, maybe this would be a nicer format for in the theme as well:
```
{
hierarchy1: {}
hierarchy2: {}
hierarchy3: {}
hierarchy4: {}
...etc
}
```

PR #101, opened 2018-07-11
Additional icon set

### LINKED ISSUE #96: Additional icon set
We currently support 2 icon sets with the `Icon` component. "small" and "large" (18x18 and 60x60 respectively). We would like to add a third one which has a canvas of 12x12. The props need to be expanded to support this third set as well.

`small | large` could become `small | medium | large`.

The svg files are inline with a webpack loader and can be imported. (we import them with require() in `https://github.com/MyOnlineStore/bricks/blob/master/src/components/Icon/types.ts` for convenience because it gives you instant assignment)

hborghols:
### This PR:

resolves #96 

**Adds** ✨
- A bunch of new icons!

**Changes** 🌀
- Icons can now be displayed in 'small' (12x12), 'medium' (18x18) or 'large' (60x60) formats

**Removes** 👋
- The following icons were removed:
_cash-register-large
checkmark-circle-large
chevron-circle-down-large
chevron-circle-left-large
chevron-circle-right-large
chevron-circle-up-large
chevron-down-large
chevron-left-large
chevron-right-large
chevron-up-large
close-large
cross-circle-large_


src/components/Icon/index.tsx
```diff
@@ -12,15 +12,25 @@ type SmallPropsType = BasePropsType & {
...
     size: 'large';
     icon: keyof typeof LargeIcons;
 };
 
-type PropsType = SmallPropsType | LargePropsType;
+type PropsType = SmallPropsType | MediumPropsType | LargePropsType;
 
 const Icon: StatelessComponent<PropsType> = (props): JSX.Element => {
-    const icon = props.size === 'large' ? LargeIcons[props.icon] : SmallIcons[props.icon];
+    const icon =
```
CarloPalinckx: maybe prefer a `switch` here for readability.

PR #114, opened 2018-07-23
Added Select

CarloPalinckx:
### This PR:

- [x] needs tests
- [x] need to look at the open animation, placeholder has some weird movement when opening.

**Adds** ✨
- A filterable `Select` component
- A `maxHeight` prop to `Box`

**Changes** 🌀
- Inline behavior for `Box` it now renders a `span` when `inline` is set to true.
- Fixes a bug where `autoHideScrollBar` on `ScrollBox` would be ignored.
- Confines `Contrast`'s theme merging behavior to construction instead of possibly every render. This should increase its performance.

**Removes** 👋
- Tslint ordered import rules. If we really care deeply about import ordening we can run the tslint fixer to sort them for us occasionally.
- `Box`'s test snapshot. These were way too brittle and now just assert the style rule applied.


package.json
```diff
@@ -86,6 +85,7 @@
     "lint-fix": "tslint --fix --project ./config/typescript/tsconfig.json --config ./config/typescript/tslint.json ./src/**/*.{ts,tsx}",
     "lint": "tslint --project ./config/typescript/tsconfig.json --config ./config/typescript/tslint.json ./src/**/*.{ts,tsx}",
     "watch": "start-storybook -p 9001 -c ./config/storybook",
+    "dev": "npm run watch",
```
woutsluiter: Why is this added?

package.json
```diff
@@ -19,7 +19,6 @@
     "@types/react": "^16.3.11",
     "@types/react-dom": "^16.0.5",
     "deepmerge": "^2.1.0",
-    "react-popper": "^1.0.0-beta.6",
     "resize-observer-polyfill": "^1.5.0",
```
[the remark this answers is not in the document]
CarloPalinckx: Something probably went wrong with rebasing, ill revert this 👍

package.json
CarloPalinckx: just an alias, because the conventional name for starting a dev server is `dev`

CarloPalinckx:
> when you set a predefined/selected value, the idle state shows the correct label ("Bar F"), but when you open the Select it shows the value ("F"). I would expect it to be empty since I didn't type anything yet.

could you shoot a gif of this? Can't seem to reproduce it.

> The results are always one cycle behind the typed value
used the state input value before updating it, its using event value directly now. 😅

src/components/Select/style.tsx
```diff
@@ -0,0 +1,109 @@
...
+type WindowProps = {
+    isOpen: boolean;
+};
+
+const StyledWindow = withProps<WindowProps, HTMLDivElement>(styled.div)`
+    transition: border .3s, width .3s, left .3s;
+    position: absolute;
+    max-height: 240px;
+    overflow: hidden;
+    width: ${({ isOpen }): string => (isOpen ? '100%' : 'calc(100% - 4px)')};
```
woutsluiter: The whole search field seems to resize 1px. This makes the whole transition a bit jerky. I think it would feel a lot smoother if the search field wouldn't resize/move at all.

woutsluiter:
Lol FF
image

CarloPalinckx:
The icon is just shy in a new environment 😄

woutsluiter:
>could you shoot a gif of this? Can't seem to reproduce it.
It occurs when you set the `state.value` in the story to "F" for example.

foo

src/components/Select/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+            this.state.optionPointer < this.state.filteredOption.length ? this.state.optionPointer + 1 : 0;
+
+        this.setState({ optionPointer });
+    };
+
+    private cycleDown = (): void => {
+        const optionPointer =
+            this.state.optionPointer > 0 ? this.state.optionPointer - 1 : this.state.filteredOption.length;
+
+        this.setState({ optionPointer });
```
woutsluiter: I'm getting these TS errors:
`Type argument cannot be inferred from usage`

src/components/Select/index.tsx
```diff
@@ -0,0 +1,221 @@
...
+    public componentDidMount(): void {
+        document.addEventListener('mousedown', this.handleClickOutside);
+    }
+
+    public componentWillUnmount(): void {
+        document.removeEventListener('mousedown', this.handleClickOutside);
+    }
+
+    public handleClickOutside = (event: MouseEvent): void => {
+        if (this.wrapperRef !== null && !this.wrapperRef.contains(event.target as Node)) {
```
woutsluiter: `Binary operation argument type HTMLDivElement is not compatible with type null`

PR #118, opened 2018-07-25
Popover offset on both axes

### LINKED ISSUE #116: Offset on both axes on Popover
The Popover now has an offset prop to set an offset on one axis. Because of positioning issues it would be nice to be able to set an offset on the other axis aswel.

hborghols:
### This PR:

proposed release: `minor`

resolves #116 

**Changes** 🌀
- Popover offset can now be set for both the horizontal and the vertical axes, instead of only for the vertical axis. This is implemented using `offset` and `distance` props. Distance specifies the distance between toggle and popover while offset specifies the offset of the popover body relative to the popover arrow.

### DIFF
--- a/src/components/Popover/index.tsx
+++ b/src/components/Popover/index.tsx
@@ -10,11 +10,25 @@ type PropsType = {
     isOpen: boolean;
     fixed?: boolean;
     offset?: number;
+    distance?: number;
     stretch?: boolean;
     renderContent(): JSX.Element | string;
 };
 
 const Popover: StatelessComponent<PropsType> = (props): JSX.Element => {
+    const mapOffset = (props: PropsType): string => {
+        switch (true) {
+            case props.offset === undefined && props.distance === undefined:
+                return '0, 16px';
+            case props.offset !== undefined && props.distance === undefined:
+                return `${props.offset}px, 16px`;
+            case props.offset === undefined && props.distance !== undefined:
+                return `0, ${props.distance}px`;
+            default:
+                return `${props.offset}px, ${props.distance}px`;
+        }
+    };
+
     return (
         <>
             <Manager>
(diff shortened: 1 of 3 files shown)


src/components/Popover/index.tsx
```diff
@@ -4,16 +4,31 @@ import TransitionAnimation from '../TransitionAnimation';
 import { PopoverAnchor, PopoverArrow, PopoverBackground, PopoverContent, PopoverWindow } from './style';
 
 type PlacementType = PopperChildrenProps['placement'];
+type OffsetType = {
+    horizontal?: number;
```
woutsluiter: Not sure if this is always a horizontal axis. When you set the placement to `right` or `left` these might be rotated.

woutsluiter: you could also rename the `offset` props to `distance` and use `offset` for the new axis

src/components/Popover/index.tsx
```diff
@@ -31,7 +43,10 @@ const Popover: StatelessComponent<PropsType> = (props): JSX.Element => {
                         placement={props.placement !== undefined ? props.placement : 'bottom'}
                         modifiers={{
                             offset: {
-                                offset: props.offset !== undefined ? `0, ${props.offset}px` : '0, 16px',
+                                offset:
+                                    props.offset !== undefined || props.distance !== undefined
+                                        ? mapOffset(props)
+                                        : '0, 16px',
```
woutsluiter: You could include this in the `mapOffset` aswel.

src/components/Popover/index.tsx
```diff
@@ -10,10 +10,22 @@ type PropsType = {
     isOpen: boolean;
     fixed?: boolean;
     offset?: number;
+    distance?: number;
     stretch?: boolean;
     renderContent(): JSX.Element | string;
 };
 
+const mapOffset = (props: PropsType): string => {
```
woutsluiter: This should be inside Popover

PR #125, opened 2018-07-30
Added fixed width to modal

CarloPalinckx:
### This PR:

proposed release: `patch`

**Adds** ✨
- Fixed width to the modal to allow for percentual styling inside of it.

### DIFF
--- a/src/components/Modal/story.tsx
+++ b/src/components/Modal/story.tsx
@@ -5,6 +5,7 @@ import Modal from '.';
 import Button from '../Button';
 import ButtonGroup from '../ButtonGroup';
 import Text from '../Text';
+import Box from '../Box';
 
 const demoContent = `
 Lorem ipsum dolor sit amet, consectetur adipiscing elit. Etiam ut mi ligula. Phasellus tellus nulla,
--- a/src/components/Modal/story.tsx
+++ b/src/components/Modal/story.tsx
@@ -5,7 +5,6 @@ import Modal from '.';
 import Button from '../Button';
 import ButtonGroup from '../ButtonGroup';
 import Text from '../Text';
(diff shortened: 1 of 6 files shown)


src/components/Modal/style.tsx
```diff
@@ -19,25 +19,28 @@ const StyledModalWrapper = withProps<ModalWrapperPropsType>(styled.div)`
...
+        max-height: 100%;
+    }
 `;
 
 const StyledModal = styled.div`
-    margin: auto;
-    max-width: 600px;
-    width: calc(100% - 24px);
+    max-width: 100%;
+    width: 791px;
```
woutsluiter: Peculiar value 😄

src/components/Modal/style.tsx
```diff
@@ -19,25 +19,28 @@ const StyledModalWrapper = withProps<ModalWrapperPropsType>(styled.div)`
...
     flex-grow: 1;
```
nicokoelewijn: flex-grow can be removed

src/components/Modal/style.tsx
CarloPalinckx: Oops . this should be 792 because ⤵️

CarloPalinckx: screen shot 2018-07-30 at 09 46 52

src/components/Modal/story.tsx
```diff
@@ -5,6 +5,7 @@ import Modal from '.';
 import Button from '../Button';
 import ButtonGroup from '../ButtonGroup';
 import Text from '../Text';
+import Box from '../Box';
```
woutsluiter: Unused

PR #134, opened 2018-07-31
Added custom rendering story to select

CarloPalinckx:
### This PR:

proposed release: `patch`

**Adds** ✨
- A custom rendering story to `Select`
- An `isSelected` property on the `renderOption` callback's first argument on the `Select`

**Breaking changes** 🔥
- The `renderOption` callback now receives a second argument `provided` which contains `isSelected`. You are now required to provide your own way of showing that an option is selected.


src/components/Select/index.tsx
```diff
@@ -195,29 +199,39 @@ class Select<GenericOption extends OptionBase> extends Component<PropsType<Gener
...
+                                                onMouseEnter={(): void => this.cycleTo(index)}
+                                                onClick={(): void => {
+                                                    this.handleChange(option.value);
+                                                }}
+                                            >
+                                                <Text inline descriptive={option.value === this.props.value}>
+                                                    {isSelected &&
+                                                        this.props.renderOption === undefined && (
+                                                            <Box margin={trbl(0, 6, 0, 0)} inline>
+                                                                <Icon size="small" icon="checkmark" />
```
hborghols: This doesn't render the same way as before, the checkmark and label end up on seperate lines rather than a checkmark in front of the label.

hborghols: Selecting options after an initial selection appears to be broken in chrome and safari

custom-select-broken-chrome
default-select-broken-chrome

CarloPalinckx:
closed in favor of #206

PR #138, opened 2018-08-01
button-icon

nicokoelewijn:
### This PR:

proposed release: patch

**Checklist** 🛡

**Adds** ✨
- Button icon can now be defined with props.

### DIFF
--- a/src/components/Button/story.tsx
+++ b/src/components/Button/story.tsx
@@ -8,6 +8,11 @@ import ButtonGroup from '../ButtonGroup';
 import Contrast from '../Contrast';
 import Icon from '../Icon';
 import Box from '../Box';
+import { MediumIcons } from '../Icon/types';
+
+/* tslint:disable */
+const mediumIconKeys = Object.keys(MediumIcons).filter(key => MediumIcons[key as any].match('<svg'));
+/* tslint:enable */
 
 storiesOf('Button', module)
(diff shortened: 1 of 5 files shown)


src/components/Button/story.tsx
```diff
@@ -37,19 +39,16 @@ storiesOf('Button', module)
...
-                title={text('title', 'Click me')}
+                title={text('title', 'Press any key to continue')}
                 disabled={boolean('disabled', false)}
                 flat={boolean('flat', false)}
-                action={(): void => {
-                    /* */
-                }}
+                icon={select('Icon', mediumIconKeys, 'keyboard') as PropsType['icon']}
+                iconAlign={select('Align icon', ['left', 'right'], 'right') as PropsType['iconAlign']}
                 compact
```
hborghols: In the story for button with text this is toggleable, would be consistent to do it here as well. Also because you can toggle the `flat` prop as well.

src/components/Button/style.tsx
```diff
@@ -77,6 +77,7 @@ type ButtonPropsType = {
     compact?: PropsType['compact'];
     flat?: PropsType['flat'];
     variant: PropsType['variant'];
+    icon?: string;
```
hborghols: This can be `PropsType['icon']` to be consistent with the other props.

PR #140, opened 2018-08-01
Update issue templates

CarloPalinckx:


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,17 @@
+---
+name: Feature request
+about: Suggest an idea for this project
+
+---
+
+**Is your feature request related to a problem? Please describe.**
```
hborghols: Picking some nits: there is a period here at the end of the line but not in the other headings. Would be nice if this would be consistent across the file and even nicer if it were consistent across both templates. So maybe we need some nice emoji's here as well :smile:.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,17 @@
...
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]
```
hborghols: I would add quotes and a semicolon. Ex: "I'm always frustrated when [...]".

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,17 @@
...
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
```
hborghols: This is maybe a bit vague. Maybe change it to: A clear and concise description of what you'd like to see added or changed.

PR #141, opened 2018-08-02
Added text skeleton

CarloPalinckx:
### This PR:

proposed release: `patch`

**Checklist** 🛡

**Adds** ✨
- A `Skeleton` component.

### DIFF
--- a/src/components/Skeleton/story.tsx
+++ b/src/components/Skeleton/story.tsx
@@ -20,7 +20,7 @@ const Demo: SFC<DemoPropsType> = ({ loading }): JSX.Element => {
     return (
         <div>
             <Text>Leverage agile frameworks to provide a robust synopsis for high level overviews.</Text>
-            <Text> Iterative approaches to corporate strategy foster collaborative</Text>
+            <Text>Iterative approaches to corporate strategy foster collaborative</Text>
             <Text>thinking to further the overall value proposition.</Text>
         </div>
     );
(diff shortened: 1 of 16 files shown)


src/components/Skeleton/story.tsx
```diff
@@ -0,0 +1,29 @@
...
+            <div>
+                <Skeleton.Text lines={number('lines', 3)} baseWidth={number('baseWidth', 80)} />
+            </div>
+        );
+    }
+
+    return (
+        <div>
+            <Text>Leverage agile frameworks to provide a robust synopsis for high level overviews.</Text>
+            <Text> Iterative approaches to corporate strategy foster collaborative</Text>
```
hborghols: Supernit: the sentence in the Text component starts with whitespace which I assume is a typo.

src/utility/calculateRandomPercentage/index.ts
```diff
@@ -0,0 +1,8 @@
+const calculateRandomPercentage = (min:number, max:number):number => {
```
nicokoelewijn: a default value of zero on min parameter would be a nice addition! less code and in most cases only the max boundary needs to be passed https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

src/components/Skeleton/test.tsx
```diff
@@ -0,0 +1,20 @@
...
+describe('Skeleton', () => {
+    it('should not crash when no baseWidth is set', () => {
+        const fn = (): void => {
+            mountWithTheme(<Skeleton.Text lines={1} />);
+        };
+
+        expect(fn).not.toThrow();
+    });
+
+    it('should not render to small placeholders', () => {
```
hborghols: I find this a bit vague. I would change it to 'should not render placeholders less than 6px wide" or something similar

PR #145, opened 2018-08-06
Added headers prop to Table

nicokoelewijn:
### This PR:

proposed release: `patch`

**Checklist** 🛡

**Adds** ✨
- an `onMount` and `onUnmount` prop to the `Checkbox`
- an `alignments` prop on the `Table` which lets you control column alignment
- a new `rows` props which supports a declarative way to define your rows/cells.
- `selectable`/`onSelection` prop on `Table`

**Breaking changes** 🔥

**Table/TableCell/TableRow**
- Table `dragEndHandler` changed to `onDragEnd`
- `TableCell` and `TableRow` are no longer exported, You can however solve most of the issues that required the custom rendering with the new `rows` prop on the `Table` which now supports `ReactNode` as item.

**Checkbock/RadioButton**
- `Checkbox` and `RadioButton`'s `changeHandler` changed to `onChange`
- Checkbox now supports a checked prop of: `boolean`|`'indeterminate'`, this might generate typescript errors when running in a strict environment when using the `Checkbox`'s `onChange` prop.


src/components/Table/index.tsx
```diff
@@ -3,10 +3,12 @@ import { DragDropContext, Droppable, DropResult } from 'react-beautiful-dnd';
 import StyledTable from './style';
 import TableRow from '../TableRow';
 import TableCell from '../TableCell';
+import Text from '../Text';
 
 type PropsType<GenericRowType, GenericCellType> = {
     children?: ReactNode;
     data?: ReadonlyArray<GenericRowType>;
+    headingData?: GenericRowType;
```
CarloPalinckx: You don't want to share the RowType that is used on the data prop here. They will never have a shared structure. Imo it would be best if this was just an array of strings for now that is rendered left to right. 

`headers?: Array<string>`

src/components/Table/index.tsx
```diff
@@ -30,6 +32,16 @@ const Table = <GenericCellType extends CellBaseType, GenericRowType extends Arra
                     <StyledTable innerRef={innerRef}>
                         <tbody>
                             {props.children !== undefined && props.children}
+                            {props.headingData !== undefined && (
```
CarloPalinckx: headers should be above the children here.

src/components/Table/index.tsx
```diff
@@ -3,12 +3,14 @@ import { DragDropContext, Droppable, DropResult } from 'react-beautiful-dnd';
 import StyledTable from './style';
 import TableRow from '../TableRow';
 import TableCell from '../TableCell';
+import Text from '../Text';
 
 type PropsType<GenericRowType, GenericCellType> = {
     children?: ReactNode;
     data?: ReadonlyArray<GenericRowType>;
+    headingData?: Array<string>;
```
CarloPalinckx: I think headers instead of headingData would be explicit enough 👍

src/components/Table/index.tsx
```diff
@@ -3,12 +3,14 @@ import { DragDropContext, Droppable, DropResult } from 'react-beautiful-dnd';
...
     draggable?: boolean;
-    dragEndHandler?(result: DropResult): void;
+    onDragEnd?(result: DropResult): void;
```
CarloPalinckx: changing this prop is a breaking change. Make sure to add this to the PR according to the PR template.

src/components/Table/Row/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+                                        onUnmount={(): void => remove(identifier)}
+                                    />
+                                );
+                            }}
+                        </SubscriptionConsumer>
+                    </Cell>
+                )}
+                {cells.map((cell, cellIndex: number) => (
+                    <Cell key={`cell-${cellIndex}`}>
+                        <Box justifyContent={alignments[cellIndex]}>
```
woutsluiter: A Box with alignment is not enough. You also need to set the `text-align`.

src/components/Table/Header/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+                                    />
+                                );
+                            }}
+                        </SubscriptionConsumer>
+                    </StyledHeader>
+                )}
+                {headers.map((header, headerIndex): JSX.Element => {
+                    return (
+                        <StyledHeader key={headerIndex}>
+                            <Box justifyContent={alignments[headerIndex]}>
```
woutsluiter: A Box with alignment is not enough. You also need to set the `text-align`.

src/components/Table/Header/style.tsx
```diff
@@ -1,10 +1,24 @@
...
+import styled, { withProps } from '../../../utility/styled';
 
-const StyledHeader = styled.th`
+type PropsType = {
+    align: 'flex-start' | 'center' | 'flex-end';
+};
+
+const StyledHeader = withProps<PropsType>(styled.th)`
     padding: 12px;
+    text-align: ${({ align }): string => {
```
woutsluiter: Wouldn't it be easier to use with `left | center | right` as props and translate that to flex properties?

PR #149, opened 2018-08-09
Reduced bundle size

### LINKED ISSUE #117: Bundle optimisation thread
The bundle size is getting pretty hefty. We should have a good look at what we are actually exporting. Because it seems way too big for what we are building (1.3mb currently).

Few thing to investigate:
- Alternatives for inlined svg's. These are now all imported once you import the Icon component.
- Look at declaring react as an external in package.json. I think we are somehow still shipping react + react-dom even though it is a peerDependency.

CarloPalinckx:
### This PR:

**Baseline**
build time: `~19792ms`
bundle size: `1.1mb actual | 2.2mb raw`

**Result**
build time: `~12225ms`
bundle size: `459.7kb actual | 949.9kb raw` 🎉 

------------------

proposed release: `major`
resolves #117

**Checklist** 🛡

**Adds** ✨
- Extra bundle visualisation placed in /reports/webpack after each `npm run build`

**Changes** 🌀
- moves `mountWithTheme` and `shallowWithTheme` from `utility/styled/index.tsx` to `utility/styled/testing.tsx`
- decreases bundle size

**Breaking Changes** 🔥
- bricks now requires `react-dom` and `styled-components` as peerDependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,15 +16,11 @@
     "components"
   ],
   "dependencies": {
-    "@types/react": "^16.3.11",
-    "@types/react-dom": "^16.0.5",
     "deepmerge": "^2.1.0",
     "react-beautiful-dnd": "^8.0.3",
     "react-popper": "^1.0.0-beta.6",
     "resize-observer-polyfill": "^1.5.0",
-    "simplebar": "^2.6.1",
-    "styled-components": "^3.2.3",
-    "uuid": "^3.2.1"
+    "simplebar": "^2.6.1"
   },
   "devDependencies": {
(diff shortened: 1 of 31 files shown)


package.json
```diff
@@ -16,70 +16,75 @@
     "components"
   ],
   "dependencies": {
-    "@types/react": "^16.3.11",
-    "@types/react-dom": "^16.0.5",
+    "babel-loader": "^7.1.5",
```
nicokoelewijn: Is this plugin in use?

PR #156, opened 2018-08-15
create FormRow component

nicokoelewijn:
### This PR:

proposed release: `patch`

**Checklist** 🛡

**Adds** ✨
- Feature `FormRow component`


src/components/FormRow/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+import BreakpointProvider from '../BreakpointProvider';
+import Box from '../Box';
+import trbl from '../../utility/trbl';
+
+type StateType = {
+    focus: boolean;
+};
+
+type PropsType = {
+    rowLabel: ReactNode;
```
CarloPalinckx: the `row` part of these props are pretty much implied since it is a `FormRow` component. I think label would suffice here.

src/components/FormRow/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    rowValue: JSX.Element;
```
CarloPalinckx: I don't think `value` is the correct term here, it's a bit too ambiguous imo. Maybe `field` is a better term here.

src/components/FormRow/story.tsx
```diff
@@ -0,0 +1,66 @@
...
+import React from 'react';
+import FormRow from '.';
+import RadioButton from '../RadioButton';
+import RadioButtonGroup from '../RadioButtonGroup';
+import Button from '../Button';
+import Text from '../Text';
+import Box from '../Box';
+import TextField from '../TextField';
+
+storiesOf('FormRow', module).add('Default', () => (
```
CarloPalinckx: Maybe you could improve this story by creating a realistic form to show proper usage of the component (with reasonable labels).

src/components/FormRow/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+                    return (
+                        <Box
+                            margin={trbl(0, 0, 12, 0)}
+                            direction={breakpoint === 'small' ? 'column' : 'row'}
+                            grow={0}
+                            justifyContent="flex-start"
+                            alignItems="stretch"
+                        >
+                            <StyledRowDescription small={breakpoint === 'small'}>
```
CarloPalinckx: You are mixing language here, I think the internals would be clearer if they matched the language used in the props.  Since there is not really a reason not to.

src/components/FormRow/story.tsx
CarloPalinckx: You could also check with design which `Text` style should be used with the labels.

src/components/FormRow/story.tsx
```diff
@@ -0,0 +1,66 @@
...
+                    <Button title="" variant="primary" flat compact icon="gear" />
+                </Box>
+            }
+        />
+
+        <FormRow
+            rowLabel={<Text>That's a radiobutton</Text>}
+            rowValue={
+                <RadioButtonGroup>
+                    <RadioButton changeHandler={(): void => undefined} label="foobar" value="foo1" name="bar" checked />
```
CarloPalinckx: Maybe you could also keep some state in the demo to make the radiobuttons clickable (like in the story of `RadioButtonGroup`)

src/components/FormRow/style.tsx
```diff
@@ -0,0 +1,23 @@
...
+width: ${({ small }): string => (small ? '120px' : '156px')};
+margin-bottom: ${({ small }): string => (small ? '6px' : '12px')};
+flex-grow: 0;
+flex-shrink: 0;
+display: flex;`;
+
+const StyledRowValue = styled.div`
+    display: flex;
+    width: 100%;
+    margin-bottom: 12px;
```
CarloPalinckx: maybe you could only apply this bottom margin with the `small` breakpoint because the "visual" margin is now a product of value margin + row margin, which makes it a little less clear imo.

CarloPalinckx: Same with the "description"

PR #158, opened 2018-08-17
Updated TextField

nicokoelewijn:
### This PR:

proposed release: `patch`

**Checklist** 🛡

**Changes** 🌀
- Component `TextField` has new design and the label is removed.

**Breaking changes** 🔥
- `label` is removed
- `changeHandler` is now `onChange`

### DIFF
--- a/src/components/TextField/style.tsx
+++ b/src/components/TextField/style.tsx
@@ -57,14 +57,12 @@ const StyledAffixWrapper = styled.div`
 const StyledPrefix = styled.span`
     max-width: 100%;
     white-space: nowrap;
-    overflow: hidden;
     text-overflow: ellipsis;
 `;
 
 const StyledSuffix = styled.span`
     max-width: 100%;
     white-space: nowrap;
-    overflow: hidden;
     text-overflow: ellipsis;
 `;
 
@@ -78,7 +76,6 @@ type WrapperProps = {
 };
 
 const StyledWrapper = withProps<WrapperProps, HTMLDivElement>(styled.div)`
-    margin-top: 1em;
     transition: border-color 100ms, box-shadow 100ms;
     border: solid 1px ${({ focus, theme }): string =>
         focus ? theme.TextField.focus.borderColor : theme.TextField.idle.common.borderColor};
@@ -90,10 +87,13 @@ const StyledWrapper = withProps<WrapperProps, HTMLDivElement>(styled.div)`
     position: relative;
     cursor: text;
     margin-bottom: 9px;
+    overflow: hidden;
+
     ${({ feedback, theme, focus }): string =>
         feedback !== undefined && feedback.severity !== 'info' && focus === false
             ? `border: solid 1px ${theme.Text.severity[feedback.severity].color};`
             : ''};
+
     * {
         cursor: text;
     }
--- a/src/components/TextField/style.tsx
+++ b/src/components/TextField/style.tsx
@@ -49,28 +49,31 @@ const StyledInput = withProps<WrapperProps, HTMLInputElement>(styled.input)`
     outline: none;
 `;
 
-const StyledPrefix = styled.div`
+const StyledFixWrapper = styled.div`
     display: flex;
-    border-radius: ${({ theme }): string => theme.TextField.idle.prefix.borderRadius};
     padding: 0 12px;
     user-select: none;
     color: ${({ theme }): string => theme.TextField.idle.label.color};
     background-color: ${({ theme }): string => theme.TextField.idle.label.backgroundColor};
     align-items: center;
-    max-width: 40%;
     flex-shrink: 0;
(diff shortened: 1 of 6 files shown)


src/components/TextField/style.tsx
```diff
@@ -13,6 +13,16 @@ type TextFieldThemeType = {
             fontFamily: string;
             color: string;
         };
+        label: {
+            color: string;
+            backgroundColor: string;
+        };
+        prefix: {
+            borderRadius: string;
```
woutsluiter: Why not use the border-radius value from the `common` object?

src/components/TextField/style.tsx
```diff
@@ -13,6 +13,16 @@ type TextFieldThemeType = {
             fontFamily: string;
             color: string;
         };
+        label: {
```
woutsluiter: The name label is somewhat confusing since there is no label present. You could use `affix`

src/components/TextField/style.tsx
```diff
@@ -24,47 +34,46 @@ type TextFieldThemeType = {
...
-
-const StyledFloatingLabel = withProps<FloatingLabelProps, HTMLLabelElement>(styled.label)`
-    transition: top 0.2s, transform 0.2s, font-size 0.2s, color 0.2s;
-    display: block;
-    line-height: 1.5;
-    position: absolute;
-    transform-origin: top left;
-    top: 0;
-    font-size: 1em;
+const StyledFixWrapper = styled.div`
```
woutsluiter: `StyledAffixWrapper` would be nicer imo

src/components/TextField/style.tsx
```diff
@@ -77,28 +86,26 @@ type WrapperProps = {
 };
 
 const StyledWrapper = withProps<WrapperProps, HTMLDivElement>(styled.div)`
-    transition: border-color 0.3s;
+    margin-top: 1em;
+    transition: border-color 100ms, box-shadow 100ms;
     border: solid 1px ${({ focus, theme }): string =>
         focus ? theme.TextField.focus.borderColor : theme.TextField.idle.common.borderColor};
+    box-shadow: ${({ focus, theme }): string => (focus ? theme.TextField.focus.boxShadow : '')};
+    border-radius: 3px;
```
woutsluiter: This is duplicate and a fixed value instead of themed. There's already a border-radius line on L97.

src/components/TextField/style.tsx
nicokoelewijn: Removed completely since the wrapper clips this anyway.

src/components/TextField/style.tsx
nicokoelewijn: Nice catch, the line will be removed!

src/components/TextField/style.tsx
```diff
@@ -13,58 +13,57 @@ type TextFieldThemeType = {
...
-            : `
-                color: ${theme.TextField.idle.common.color};
-                transform: translate3d(0, 0.85em, 0) scale(1);
-            `;
-    }};
+const StyledPrefix = styled.span`
+    max-width: 100%;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+`;
```
woutsluiter: The overflow hidden is needed for the ellipsis

src/components/TextField/style.tsx
```diff
@@ -77,28 +76,27 @@ type WrapperProps = {
...
+    box-shadow: ${({ focus, theme }): string => (focus ? theme.TextField.focus.boxShadow : '')};
     font-size: ${({ theme }): string => theme.TextField.idle.common.fontSize};
     font-family: ${({ theme }): string => theme.TextField.idle.common.fontFamily};
     border-radius: ${({ theme }): string => theme.TextField.idle.common.borderRadius};
-    display: inline-block;
+    display: flex;
     position: relative;
-    padding: 0.3em 12px 0.4em 12px;
     cursor: text;
+    margin-bottom: 9px;
```
woutsluiter: Leftover margin

src/components/TextField/style.tsx
woutsluiter: Also, when I look in Abstract there is a border between the input and the affix:
image

src/components/TextField/style.tsx
```diff
@@ -13,58 +13,67 @@ type TextFieldThemeType = {
...
+    max-width: 100%;
+    white-space: nowrap;
+    text-overflow: ellipsis;
+    overflow: hidden;
+`;
+
+const StyledSuffix = styled.span`
+    max-width: 100%;
+    white-space: nowrap;
+    text-overflow: ellipsis;
```
woutsluiter: This is now missing an overflow hidden. And if you add it's exactly the same as the StyledPrefix, so you might aswel refactor it to one `StyledAffix`.

PR #159, opened 2018-08-20
Added Currency formatting to TextField

CarloPalinckx:
### This PR:

proposed release: `minor`

**Checklist** 🛡

**Adds** ✨
- currency formatting to `TextField`
- `onBlur` props to `TextField`
- `polyfill` as export which provides users with polyfills for bricks

**Changes** 🌀
- `npm run test` no longer run with the coverage flag, `npm run test-coverage` now does.
-  typescript to `2.8`


src/components/TextField/formatters/withCurrencyFormatting/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+        private handleChange = (value: string): void => {
+            this.setState({ value: this.parseIn(value) });
+            this.props.onChange(this.parseOut(value));
+        };
+
+        private handleBlur = (): void => {
+            this.setState({ value: this.format(this.state.value) });
+        };
+
+        public componentWillMount(): void {
```
woutsluiter: 😱

CarloPalinckx: oops i did a deprecate

src/components/TextField/formatters/withCurrencyFormatting/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        private handleChange = (value: string): void => {
+            this.setState({ value: this.parseIn(value) });
+            this.props.onChange(this.parseOut(value));
+        };
+
+        private handleBlur = (): void => {
+            this.setState({ value: this.format(this.state.value) });
+        };
```
hborghols: When one of the settings is changed (e.g., EUR to USD) this only shows after triggering a blur event. This is a bit counter-intuitive when not in a focus state imo. For example if you would enter an amount and then realize you would like to use a different currency, I would expect the currency symbol and formatting immediately after finishing my selection when the input is not focused.

src/components/TextField/formatters/withCurrencyFormatting/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        private parseIn(value: string): string {
+            const regex = new RegExp(`[^0-9${this.state.decimalSeperator}]`);
+
+            return value.replace(regex, '');
+        }
+
+        private parseOut(value: string): number {
+            const parsed = parseFloat(value.replace(this.state.decimalSeperator, '.'));
```
hborghols: Large numbers are not parsed correctly after having been parsed once. When parsing numbers > 1000, the combination of decimal seperators and thousands seperators causes rounding errors when triggering another blur event. 

For example: when entering `1360` in EUR, this is formatted as `1.360,00`. When subsequently triggering another blur event without changing the input, this is reformatted to `1,36`.

src/components/TextField/formatters/withCurrencyFormatting/test.tsx
```diff
@@ -0,0 +1,135 @@
...
+                value: '19000000.121111111',
+            },
+        });
+
+        component.find('input').simulate('blur');
+
+        expect(component.find('input').prop('value')).toEqual('19,000,000.12');
+        expect(component.find(TextField).prop('prefix')).toEqual('$');
+        expect(changeMock).toHaveBeenCalledWith(19000000.12);
+    });
```
hborghols: Maybe a test should be added to make sure input is not formatted multiple times with different results (i.e., the rounding errors case described above).

hborghols:
The field does not show prefix or suffix in safari and formatting doesn't work. Disallowed input (e.g., `aaaaaa`) is prevented correctly but no formatted output is shown.

CarloPalinckx:
> The field does not show prefix or suffix in safari, IE11 and Edge, and formatting doesn't work. 

This is correct. Bricks will now require a polyfill to run. I figured this could be a separate issue because we could provide an easy way to provide other polyfills as well. Will add an issue 👍

CarloPalinckx:
Done ✅ , #160

PR #166, opened 2018-08-22
Added disabled icon and label to toggle

hborghols:
### This PR:

proposed release: `patch`

**Checklist** 🛡

**Changes** 🌀
- Label for `Toggle` is now optional.
- `Toggle` to actually display the toggle label when set.
- Added optional `disabledIcon` prop to `Toggle` to show a lock next to a toggle in disabled state.
- Changed `Toggle` story to a stateful demo with clickable toggle.
- Styling of disabled && checked toggle state and toggle label to more closely reflect the design.

### DIFF
--- a/src/components/Toggle/index.tsx
+++ b/src/components/Toggle/index.tsx
@@ -1,6 +1,8 @@
 import React, { Component } from 'react';
 import trbl from '../../utility/trbl';
 import Box from '../Box';
+import Text from '../Text';
+import Icon from '../Icon';
 import StyledToggle, { StyledToggleSkin, StyledToggleWrapper } from './style';
 
 type StateType = {
@@ -10,9 +12,10 @@ type StateType = {
 type PropsType = {
     checked: boolean;
     disabled?: boolean;
+    disabledIcon?: boolean;
     error?: boolean;
     value: string;
-    label: string;
+    label?: string;
     name: string;
     id?: string;
     changeHandler(change: { checked: boolean }): void;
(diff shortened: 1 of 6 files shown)


src/components/Toggle/index.tsx
```diff
@@ -10,9 +12,10 @@ type StateType = {
...
     checked: boolean;
     disabled?: boolean;
+    disabledIcon?: boolean;
     error?: boolean;
     value: string;
-    label: string;
+    label?: string;
     name: string;
     id?: string;
     changeHandler(change: { checked: boolean }): void;
```
woutsluiter: Could you rename this to onChange? Kind of a new convention to maintain react props. And maybe mark it as a breaking change.

src/components/Toggle/index.tsx
```diff
@@ -61,6 +64,10 @@ class Toggle extends Component<PropsType, StateType> {
                         />
                     </StyledToggleSkin>
                 </Box>
+                <Text descriptive={this.props.disabled ? true : false}>
+                    {this.props.disabledIcon && this.props.disabled && <Icon size="medium" icon="locked" />}{' '}
```
woutsluiter: Maybe `disabledIcon` should be true by default. That way you'd have to explicitly break the pattern if needed. You could do it with default props.

src/components/Toggle/index.tsx
```diff
@@ -61,10 +71,14 @@ class Toggle extends Component<PropsType, StateType> {
                         />
                     </StyledToggleSkin>
                 </Box>
+                <Text descriptive={this.props.disabled ? true : false}>
```
woutsluiter: This could be simplified to `!!this.props.disabled`

PR #187, opened 2018-09-06
added disabled state to Select Component

robinvdv12:
### This PR:

proposed release: `patch`

**Checklist** 🛡

**Adds** ✨
- a `disabled` state to the Select Component

### DIFF
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -16,6 +16,7 @@ type OptionBase = {
 };
 
 type StateType = {
+    disabled: boolean;
     input: string;
     isOpen: boolean;
     optionPointer: number;
@@ -26,6 +27,7 @@ type PropsType<GenericOption extends OptionBase> = {
     value: string;
     options: Array<GenericOption>;
     emptyText: string;
+    disabled?: boolean;
     onChange(value: string): void;
     renderOption?(option: GenericOption): JSX.Element;
 };
@@ -39,6 +41,7 @@ class Select<GenericOption extends OptionBase> extends Component<PropsType<Gener
         this.inputRef = createRef();
 
         this.state = {
+            disabled: props.disabled !== undefined ? props.disabled : false,
             isOpen: false,
             input: props.value,
             optionPointer: -1,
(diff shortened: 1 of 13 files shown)


src/components/Select/index.tsx
```diff
@@ -16,6 +16,7 @@ type OptionBase = {
 };
 
 type StateType = {
+    disabled: boolean;
```
CarloPalinckx: this shouldn't be in the state since it doesn't have to change internally.

CarloPalinckx:
The design seems a bit off from the UX resource.

src/components/Select/story.tsx
```diff
@@ -63,6 +63,7 @@ class Demo extends Component<{}, { value: string }> {
                 emptyText={text('emptyText', 'No results')}
                 onChange={this.handleChange}
                 options={object('options', options)}
+                disabled={boolean('disabled', true)}
```
nicokoelewijn: Please move the disabled toggle above the options. Due to a bug in storybook the option textarea takes to much space, hiding the disabled toggle.

src/components/Select/style.tsx
```diff
@@ -98,13 +111,17 @@ const StyledWindow = withProps<WindowProps, HTMLDivElement>(styled.div)`
...
-const StyledInput = styled.div`
+const StyledInput = withProps<InputProps>(styled.div)`
     box-sizing: border-box;
     width: 100%;
     padding: 0 0 0 12px;
     border: solid 1px ${({ theme }): string => theme.Select.input.borderColor};
-    background: ${({ theme }): string => theme.Select.input.background};
+    background: ${({ theme, disabled }): string =>
+        disabled
+            ? '#f5f6f7 repeating-linear-gradient( -45deg, #f0f1f2, #f0f1f2 10px, transparent 10px, transparent 20px )'
```
nicokoelewijn: Please make this background theme-able

src/components/Select/story.tsx
nicokoelewijn: Also, disabled isn't the default value on any of the component's stories. Could you make this false?

PR #199, opened 2018-09-18
Solved currency input bugs

### LINKED ISSUE #186: Textfield input (currency)
### Bricks version 📦
0.1.0

### Describe the bug 🔍
When the value of the currency is edited, the cursor will be moved to the end of the input. This is likely caused by the regex that removes the dot in large numbers, this behaviour is also caused by filling in a character. 

### To Reproduce 💥
1. Go to Bricks > Textfield > currency
2. Fill in a character anywhere (not the end) in the string

1. Go to Bricks > Textfield > currency
2. fill the input with `100000`
3. blur the input
4. add a number to the string. 

### Expected behavior ✨
The cursor should stay at the right position. 
Another solution could be to remove the dot(s) on focus.

nicokoelewijn:
### This PR:

proposed release: `minor`
resolves #186 

**Checklist** 🛡

**Changes** 🌀
- Cursor no longer jumps to end when a NaN-character is pressed. 
- Input can now be disabled
- When the currency input is empty on blur, the value will be set to 0 instead of reverting to the last value.

**Breaking changes 🔥**

- `backgroundColor` properties in `theme.TextField` changed to `background`.

### DIFF
--- a/src/components/TextField/formatters/withCurrencyFormatting/index.tsx
+++ b/src/components/TextField/formatters/withCurrencyFormatting/index.tsx
@@ -105,19 +105,15 @@ const withCurrencyFormatting = (Wrapped: ComponentType<TextFieldPropsType>): Com
             this.props.onChange(this.parse('out', value));
 
             const target = event.target;
+            const selectionStart = target.selectionStart as number;
             const newInputLength = this.parse('in', value).length;
 
-            if (target.selectionStart !== null) {
-                this.setState({
-                    cursorPosition: target.selectionStart,
-                    inputLength: target.value.length,
-                });
-            }
+            this.setState({
+                inputLength: target.value.length,
+            });
 
             this.setState({ value: this.parse('in', value) }, () => {
-                this.state.inputLength === newInputLength
-                    ? (target.selectionEnd = this.state.cursorPosition)
-                    : (target.selectionEnd = this.state.cursorPosition - 1);
+                target.selectionEnd = this.state.inputLength === newInputLength ? selectionStart : selectionStart - 1;
             });
         };
 
--- a/src/components/TextField/formatters/withCurrencyFormatting/index.tsx
+++ b/src/components/TextField/formatters/withCurrencyFormatting/index.tsx
@@ -1,4 +1,4 @@
-import React, { Component, ComponentClass, ComponentType } from 'react';
+import React, { Component, ComponentClass, ComponentType, ChangeEvent } from 'react';
 import { PropsType as TextFieldPropsType } from '../../';
 
 type OmittedKeys = 'onChange' | 'value';
@@ -13,8 +13,10 @@ type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, Omit
 type StateType = {
     value: string;
     currency: string;
+    cursorPosition: number;
     currencyAlignment: 'left' | 'right';
     decimalSeperator: string;
+    inputLength: number;
 };
 
 type PartType = {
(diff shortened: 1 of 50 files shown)


src/components/TextField/formatters/withCurrencyFormatting/index.tsx
```diff
@@ -105,13 +101,32 @@ const withCurrencyFormatting = (Wrapped: ComponentType<TextFieldPropsType>): Com
...
+
+            this.setState({ value: this.parse('in', value) }, () => {
+                target.selectionEnd = this.state.inputLength === newInputLength ? selectionStart : selectionStart - 1;
+            });
         };
 
         private handleBlur = (): void => {
-            this.setState({ value: this.format(this.state.value) });
+            if (this.state.value.length !== 0) {
+                return this.setState({ value: this.format(this.state.value) });
```
CarloPalinckx: This return seems a bit misplaced. If the types of setState would change to return something builds would break, prefer an if/else here

src/components/TextField/formatters/withCurrencyFormatting/index.tsx
```diff
@@ -105,13 +101,32 @@ const withCurrencyFormatting = (Wrapped: ComponentType<TextFieldPropsType>): Com
...
+            }
```
CarloPalinckx: Nit: newline

src/components/TextField/style.tsx
```diff
@@ -31,29 +31,51 @@ type TextFieldThemeType = {
         borderColor: string;
         boxShadow: string;
     };
+    disabled: {
+        color: string;
+        background: string;
+    };
+};
+
+type affixProps = {
```
CarloPalinckx: Types and Classes should be uppercase

src/components/TextField/style.tsx
```diff
@@ -31,29 +31,51 @@ type TextFieldThemeType = {
...
 `;
 
-const StyledAffixWrapper = styled.div`
+const StyledAffixWrapper = withProps<affixProps, HTMLDivElement>(styled.div)`
     display: flex;
     padding: 0 12px;
     user-select: none;
-    color: ${({ theme }): string => theme.TextField.idle.affix.color};
-    background-color: ${({ theme }): string => theme.TextField.idle.affix.backgroundColor};
+
```
CarloPalinckx: Nit: newline

src/components/TextField/index.tsx
```diff
@@ -60,33 +64,33 @@ class TextField extends Component<PropsType, StateType> {
...
 
-        if (this.props.onBlur !== undefined) {
-            this.props.onBlur();
-        }
+        if (this.props.onBlur !== undefined) this.props.onBlur();
     };
 
     public onChange = (event: ChangeEvent<HTMLInputElement>): void => {
-        this.props.onChange(event.target.value);
+        this.props.onChange(event.target.value, event);
```
CarloPalinckx: This should probably check this.props.disabled. It shouldn't fire the callback once disabled. Check Toggle for similar behaviour.

src/components/TextField/test.tsx
```diff
@@ -11,6 +11,14 @@ describe('TextField', () => {
         expect(component.find(StyledWrapper).prop('active')).toBe(false);
     });
 
+    it('should render a disabled state', () => {
+        const component = mountWithTheme(
+            <TextField value="" disabled suffix={'$'} name="firstName" onChange={jest.fn()} />,
+        );
+
+        expect(component.find(StyledWrapper).prop('active')).toBe(false);
+    });
```
CarloPalinckx: Maybe you could add a test case to check if the onChange isn't fired when disabled.

PR #202, opened 2018-09-19
raised hoc

### LINKED ISSUE #165: HOC's for Generic styling components
There should be different depth levels, which is currently implemented with components like Raised and Tile. But these components get in the way. And nesting hell would be inevitable. 

A better solution to this problem would be to make this behaviour optional with HOC's. 

Another option would be to add this behaviour to the Box component. But the Box already has a lot of functionalities. Making this component any bigger wouldn't be a good thing. We don't want monster components. 

### Additional context 🖼
`withRaised(Box)` would add the styling and props to the Box when it's required.

nicokoelewijn:
### This PR:

proposed release: `patch`
resolves #165 

**Checklist** 🛡

**Adds** ✨
- Adds `withRaised` HOC

**Changes** 🌀
- `Toaster` now uses the withRaised HOC

**Removes** 👋
- depracted `Raised` component

### DIFF
--- a/src/components/WithRaised/index.tsx
+++ b/src/components/WithRaised/index.tsx
@@ -1,0 +1,22 @@
+import React, { ComponentType, Component } from 'react';
+import StyledWithRaised from './style';
+
+type WithLoadingProps = {
+    level: 0 | 1 | 2;
+};
+
+// tslint:disable
+const withRaised = <P extends object>(Wrapped: ComponentType<P>) =>
+    class WithLoading extends Component<P & WithLoadingProps> {
+        public render(): JSX.Element {
+            const { level, ...props } = this.props as WithLoadingProps;
+
+            return (
+                <StyledWithRaised level={level}>
+                    <Wrapped {...props} />
+                </StyledWithRaised>
+            );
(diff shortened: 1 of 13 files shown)


src/components/WithRaised/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+};
+
+// tslint:disable
+const withRaised = <P extends object>(Wrapped: ComponentType<P>) =>
+    class WithLoading extends Component<P & WithLoadingProps> {
+        public render(): JSX.Element {
+            const { level, ...props } = this.props as WithLoadingProps;
+
+            return (
+                <StyledWithRaised level={level}>
```
CarloPalinckx: This doesn't really change the current implementation if you wrap it like this 😄. You still get the extra div here because of the StyledWithRaised. You have to figure out a way to combine the actual styles only. Maybe you could have a look at the new `as` prop introduced in styled-components v4 (which is currently in beta).

CarloPalinckx:
Closing this for now since it seems inactive, feel free to let this resurface again when it is given new life.

PR #206, opened 2018-09-26
Changed renderInput to renderSelected

CarloPalinckx:
### This PR:

proposed release: `minor`

**Checklist** 🛡

**Changes** 🌀
- `renderInput` to `renderSelected`

**Breaking changes** 🔥
- Changed `Select` `renderInput` prop to `renderSelected`
- Adds `optionState` as second argument to `renderOption`. You are now required to provide your own "selected" ui for each option.


src/components/Select/index.tsx
```diff
@@ -145,8 +163,39 @@ class Select<GenericOption extends OptionBase> extends Component<PropsType<Gener
         }
     };
 
+    public componentDidUpdate(_: PropsType<GenericOption>, prevState: StateType): void {
```
woutsluiter: Why is prevProps an undescore?

woutsluiter:
I believe the odd behavior is caused by the magic mouse. Seems that due to its sensitivity, or my fat fingers, it emits scroll events right after I click. Curious about touch screens.

src/components/Select/index.tsx
CarloPalinckx: It's a convention for unused params

src/components/Select/index.tsx
```diff
@@ -32,9 +36,18 @@ type PropsType<GenericOption extends OptionBase> = {
     emptyText: string;
     disabled?: boolean;
     onChange(value: string): void;
-    renderOption?(option: GenericOption): JSX.Element;
-    renderInput?(inputOption: OptionBase, placeholder?: string): JSX.Element;
+    renderOption?(option: GenericOption, state: OptionStateType): JSX.Element;
+    renderSelected?(option: GenericOption): JSX.Element;
 };
+
+const forEachScrollable = (callback: (node: Element) => void): void => {
```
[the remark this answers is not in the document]
CarloPalinckx: TSLint? 🤯Pretty sure something is off with your IDE. Linting shouldn't report on this and typechecking passes on `dev` as well. The types seem to reflect Element too.
screenshot 2018-10-09 at 07 46 49

CarloPalinckx:
> I believe the odd behavior is caused by the magic mouse. Seems that due to its sensitivity, or my fat fingers, it emits scroll events right after I click. Curious about touch screens.

I could look into a threshold for the closing on scroll 👍

PR #207, opened 2018-09-26
Added Range Component

robinvdv12:
### This PR:

proposed release: `patch`

**Checklist** 🛡

**Adds** ✨
- `Range` Component
- `withNumberFormatting` formatter for `Textfield` Component.

**Changes** 🌀
- `Textfield` Component styling. Removed min-width.

**To-Do / Possible additions for the future** 🔮
- [x] Disabled state functionality.
- [ ] Disabled state styling.
- [ ] Tooltip on the range-slider with current value.
- [ ] Allow same values (min- and max-value are allowed to be the same)
- [x] maxValue (set a maximum value for the range, not possible to drag the slider beyond this value)
- [x] minValue (set a minimum value for the range, not possible to drag the slider under this value)
- [x] Step (increment/decrement the value by steps (default is 1))


src/components/TextField/index.tsx
```diff
@@ -89,7 +92,7 @@ class TextField extends Component<PropsType, StateType> {
                         </StyledAffixWrapper>
                     )}
                     <StyledInput
-                        type="text"
+                        type={this.props.type}
```
CarloPalinckx: This should default to 'text' when undefined

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -566,6 +566,20 @@ const theme: ThemeType = {
             boxShadow: '0 6px 48px 0 rgba(33,37,43,0.30)',
         },
     },
+    Range: {
+        track: {
+            background: silver.base,
+            border: `solid 1px ${silver.darker4}`,
+        },
+        trackActive: {
```
CarloPalinckx: You might want to nest this inside track. Like so:

```js
track: {
    common: {
        background: silver.base,
        border: `solid 1px ${silver.darker4}`,
    },
    active: {
        background: green.base,
        border: `solid 1px ${green.darker2}`,
    },
}
```

src/components/Range/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+            });
+            if (this.props.onChange !== undefined) this.props.onChange({ ...inputValues, max });
+        } else {
+            this.setState({ error: { ...error, max: true } });
+        }
+    };
+
+    private onBlurMinimumValue = (): void => {
+        const { min, max } = this.state.inputValues;
+        let newValues;
```
CarloPalinckx: You could also turn this into a const if you wanted to. This would prevent someone from mutating it later on I guess. It is a bit more verbose tho.

```js
const newValues = () => {
    if (min < this.props.minLimit) return { ...this.state.inputValues, min: this.props.minLimit };
    if (min >= max) return { ...this.state.inputValues, min: this.getMaxLowValue() };

    return { ...this.state.inputValues, min };
}();
```

src/components/Range/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+            error: { ...this.state.error, max: false },
+        });
+        if (this.props.onChange !== undefined) this.props.onChange(inputValues);
+    };
+
+    private isWithinRange = (min: number, max: number, value: number): boolean => {
+        return value <= max && value >= min;
+    };
+
+    private getMaxLowValue = (): number => {
```
CarloPalinckx: same here

src/components/Range/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+        return this.state.inputValues.max - 1;
+    };
+
+    private getMinHighValue = (): number => {
```
CarloPalinckx: same here

src/components/Range/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+        this.setState({
+            inputValues,
+            rangeValues: inputValues,
+            inputFocus: false,
+            error: { ...this.state.error, max: false },
+        });
+        if (this.props.onChange !== undefined) this.props.onChange(inputValues);
+    };
+
+    private isWithinRange = (min: number, max: number, value: number): boolean => {
```
CarloPalinckx: this doesn't have to be an arrow function since you are not passing it to a component directly. You can do: 

```js
private getMaxLowValue():number {
}
```

here. This prevents unnecessary binds.

src/components/Range/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+        return this.state.inputValues.min + 1;
+    };
+
+    public render(): JSX.Element {
+        return (
+            <Box padding={trbl(12)} direction="column">
+                <Box justifyContent="space-between">
+                    <Box width="125px">
+                        <TextField.Number
+                            feedback={this.state.error.min ? { severity: 'error', message: '' } : undefined}
```
CarloPalinckx: Seems like the focus state is still green with an error severity 😅could you change the theming for this? Also the icon is supposed to be under the field instead of next to it. This looks like a bug as well.

src/components/Range/DontTest.story.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { boolean, number } from '@storybook/addon-knobs/react';
+import { storiesOf } from '@storybook/react';
+import Range from '.';
+
+storiesOf('Range', module).add('Default', () => {
+    return (
+        <Range
+            value={{ min: 3, max: 5 }}
```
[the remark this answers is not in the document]
nicokoelewijn: This would cause a rather big story and the settings can be changed without props.

src/components/Range/index.tsx
nicokoelewijn: This is default behaviour of the `Inputfield`, these bugs can be addressed in another branch.

src/components/Range/DontTest.story.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
```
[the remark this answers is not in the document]
nicokoelewijn: has been updated!

src/components/Range/index.tsx
```diff
@@ -0,0 +1,170 @@
...
+    private onChangeMaximumValue = (max: number): void => {
+        const { inputValues, error } = this.state;
+        this.setState({ inputValues: { ...inputValues, max } });
+
+        if (this.isWithinRange(this.getMinHighValue(), this.props.maxLimit, max)) {
+            this.setState({
+                rangeValues: { ...inputValues, max },
+                error: { ...error, max: false },
+            });
+            if (this.props.onChange !== undefined) this.props.onChange({ ...inputValues, max });
```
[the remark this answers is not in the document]
nicokoelewijn: Can't, Prettier disagrees.

PR #209, opened 2018-09-26
Added name attribute to inputfield and removed unused active prop

### LINKED ISSUE #204: Missing identifiers on Textfield - Name property
Styled components allow attributes to be set on the created elements. This can easily be added.

Not many alternatives here, it's a small change.

### Additional context 🖼
according to the internet it should be something like this: 
```
export default Button.extend.attrs({
    type: 'submit'
})``;
```

nicokoelewijn:
### This PR:

proposed release: `minor`
resolves #204 

**Checklist** 🛡

**Adds** ✨
- `name` attribute added to `Textfield`

**Removes** 👋
- `active` prop removed

**Breaking changes** 🔥
- `active` removed from Textfield in theme


src/components/TextField/test.tsx
```diff
@@ -5,50 +5,31 @@ import { mountWithTheme } from '../../utility/styled/testing';
...
-
-    it('should render a disabled state', () => {
-        const component = mountWithTheme(
-            <TextField value="" disabled suffix={'$'} name="firstName" onChange={jest.fn()} />,
-        );
+    it('should render an active state', () => {
+        const component = mountWithTheme(<TextField value="Jane" name="firstName" onChange={jest.fn()} />);
 
-        expect(component.find(StyledWrapper).prop('active')).toBe(false);
+        expect(component.find(StyledInput).prop('value')).toBe('Jane');
```
CarloPalinckx: Maybe you could change the test name here? value === Jane doesn't really prove that an active state is rendered I guess

nicokoelewijn: Test didn't add anything. Not even coverage, therefore it has been removed.

CarloPalinckx:
This pr only removes internal references to 'active' which means that it shouldn't be breaking 🙂. It would be breaking if you remove the theme key. Maybe you could clean that up as well?

Issue #215, opened 2018-10-08
Buttons Buttons Buttons. CTA-Button expansion.


nicokoelewijn:
### Is your feature request related to a problem? Please describe ⚡️
The button component needs a new version. The CTA button. This button is primarily used in emails and other use cases conducted by marketing. The problem is that the button component is already rather big. It's hard to scale this component to another variant.

### Describe the solution you'd like ✨
To solve this problem it would be a good idea to split this monster button into a couple of buttons: `BareButton`, `Button`, `CtaButton` and `FlatButton`. 

**BareButton**
The BareButton will be a lower level button containing only the logic of a button like the handlers like hover, click etc. 

**Button**
Button is the button pretty much the way we know it right now, except it no longer has the flat variant. 
Also it inherits the logic from the BareButton.

**CtaButton**
CtaButton is like the Button but with different styling. 
It inherits the logic from the BareButton.

**FlatButton**
FlatButton is the compact Button without a background. 
Again it inherits the BareButton logic. 

This way the code becomes more manageable and easy to expand while the redundancy is kept to a minimum thanks to the BareButton. 

### Describe alternatives you've considered 🎡
Another way would be to combine the CtaButton with the default button, but this would cause a lot of unlikely to be used combinations and code would still be messy imo.

hborghols:
While I agree that adding the cta button to the default button might make it too big of a monster, Having four different components with the same functionality also seems a bit weird. Especially if 'only' the styling is different. What does this CtaButton look like? Will it need to have all the same variants as the default button?

hborghols:
+ Why is it necessary to have a different looking button? What's wrong with the current buttons?

nicokoelewijn:
Well, there would be three different buttons in direct use. The BareButton would be used for all the business logic, but never used directly. This structure makes it easier to create variants in which (only) the styling may vary. I think the current buttons are awesome, but marketing needs them that's why they are necessary. 
Also marketing will be interacting with Bricks directly so splitting the default and cta button would give them the opportunity to tweak the button but not create an undesired combination with knobs from both button variants. And while we are at it splitting the button into multiple buttons, i do think the flat button could be it's own button.

CarloPalinckx:
While I totally agree that we should properly seperate them codewise. I don't think that we should export them as different components because that gets confusing pretty quick imho. We should rather invest tiem in finding an elegant way to extend upon a common base for the button.

CarloPalinckx:
[...]

PR #220, opened 2018-10-15
Button refactor

nicokoelewijn:
### This PR:

The Button has been refactored, the big matrix of combinations has been reduced and all is well again in the world.
From now on, you can still use the `<Button />` component as you used to, mostly. But If you want to use the flat version of the button, you should use `<Button.Plain />`. Also the Button family has a new addition! the `<Button.Cta />`. What does it do? Exactly the same as the default Button, but with styling. Oh no! that probably means redundant code! Nope that's where you're wrong. All the button variants inherit their functionality from the master button, the `<BareButton />`. So if that button is so awesome, shouldn't we use that one directly? No! it's really ugly because it has no styling, only logic. Just stick with `<Button />` , `<Button.Plain />` and `<Button.Cta />`

resolves: #215 

proposed release: [`minor`]

**Checklist** 🛡

**Adds** ✨
- <Button.Cta />

**Changes** 🌀
- Flat variant of `Button` is now `<Button.Plain />`
- A few linting errors have been resolved (see my comments in files changed)
- action prop is deprecated. it has been replaced with onClick.

**Breaking changes** 🔥
- Yup it's breaking. the flat prop is gone, so these will have to be replaced with `Button.Plain`


src/components/Select/style.tsx
```diff
@@ -81,13 +81,13 @@ const StyledWrapper = withProps<WrapperProps, HTMLDivElement>(styled.div)`
     }
 
     ${({ theme, isDisabled, isOpen }): string => {
-        return isDisabled === true || isOpen === true
-            ? ''
-            : `&:focus {
+        return !isOpen || !isDisabled
```
nicokoelewijn: solved: `ERROR: /.../bricks/src/components/Select/style.tsx[84, 16]: This expression is unnecessarily compared to a boolean. Just negate it.`

CarloPalinckx:
Thanks for documenting this pr so well! 🙌

hborghols:
`disabled` and `compact` props are not working on the `Default - link` story

hborghols:
On the `Cta` button the box-shadow does not disappear when the button is active. This is inconsistent with the other buttons.

hborghols:
And indeed major props for the documentation on this PR! Love it.

hborghols:
> Yup it's breaking. the compact prop is gone, so these will have to be replaced with Button.Plain

I see the compact prop still exists on the default button. So this comment is only relevant for the `Cta` and `Plain` buttons or is the compact prop still there by accident?

nicokoelewijn:
> > Yup it's breaking. the compact prop is gone, so these will have to be replaced with Button.Plain
> 
> I see the compact prop still exists on the default button. So is this comment only relevant for the `Cta` and `Plain` buttons or is the compact prop still there by accident?

Ah yes that was a typo. I ment the flat prop, not the compact. this has been edited in description.

src/components/Button/story.tsx
```diff
@@ -27,27 +29,52 @@ storiesOf('Button', module)
...
+            />
+        );
+    })
+    .add('With an icon', () => {
+        return (
+            <Button.Cta
+                title={text('title', 'Press any key to continue')}
+                variant={select('variant', ['primary', 'secondary'], 'primary') as CtaPropsType['variant']}
                 icon={select('Icon', mediumIconKeys, 'keyboard') as PropsType['icon']}
                 iconAlign={select('Align icon', ['left', 'right'], 'right') as PropsType['iconAlign']}
```
CarloPalinckx: Maybe you could default this to 'left' since that is the recommended alignment?

src/components/Button/index.tsx
```diff
@@ -12,85 +12,21 @@ type PropsType = {
...
-            )}
-    </>
-);
-
-const Button: SFC<PropsType> = (props): JSX.Element => {
-    const isLink = props.href !== undefined;
+class Button extends Component<PropsType> {
+    public static Cta: WithStyledCtaType = StyledCta;
+    public static Plain: WithStyledPlainType = StyledPlain;
+    public static Default: WithStyledDefaultType = StyledDefault;
```
CarloPalinckx: I don't think this has to be exported. We this could just be rendered in the render method  when no other static is used imo

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -64,265 +64,166 @@ const theme: ThemeType = {
...
+            },
+        },
+        Cta: {
+            common: {
+                borderRadius: '24px',
+                boxShadow: '0 -2px 0 rgba(0,0,0,0.2) inset',
+            },
+            primary: {
+                color: silver.lighter1,
+                backgroundColor: '#7BCE75',
```
hborghols: Are these colors intentionally different from the regular button colors? If so, they should probably be incorporated in the greens rather than be a hardcoded hex. Otherwise they should be made consistent with the primary button.

nicokoelewijn: Yes, the colors are based on this design: https://app.goabstract.com/projects/d124a180-8bf4-11e8-a531-213ea9a5a58d/branches/master/commits/98becc014130caf1f909c84a555d17f19ce6df63/files/2FD5DB48-6FC3-43E5-9C6F-E4AF3D0BB9E2/layers/7F6C8714-632F-4DF3-B422-8B9C25781822

CarloPalinckx:
screenshot 2018-10-19 at 12 59 40
This story seems to be aligning the icon a bit weirdly.

src/components/Button/stylePlain.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+            &:hover {
+                transform: scale(1.1);
+                ${variant === 'destructive' ? `color: ${theme.Button.Plain.destructive.color}` : ''};
+            }
+
+            &:active {
+                transform: translateY(2px);
+            }
+
```
woutsluiter: Could you add specific focus styles here to prevent unwanted inherited styles? I have this problem now with the close button in the Modal in the monolith.

nicokoelewijn: Nice catch, will discuss this with design and adres the issue in another branch

CarloPalinckx:
The link button is fullwidth 🤐

src/components/Button/test.tsx
```diff
@@ -14,24 +14,19 @@ describe('Button', () => {
...
+        expect(clickMock).not.toHaveBeenCalled();
     });
 
     it('should call the passed action on click', () => {
         const clickMock = jest.fn();
 
-        const component = mountWithTheme(
-            <Button title="button title" color="#f00" variant="primary" action={clickMock} />,
-        );
+        const component = mountWithTheme(<Button title="button2 title" variant="primary" onClick={clickMock} />);
```
CarloPalinckx:
```suggestion
        const component = mountWithTheme(<Button title="button title" variant="primary" onClick={clickMock} />);
```

src/components/Button/test.tsx
```diff
@@ -54,11 +49,17 @@ describe('Button', () => {
...
-    it('should no-op on click without an action', () => {
+    it('should no-op on click without an onClick', () => {
         const fn = (): void => {
             const component = mount(
                 <MosTheme>
-                    <Button icon={'checkmark'} title="button title" variant="primary" action={undefined} />
+                    <Button
+                        icon={'checkmark'}
+                        iconAlign="right"
+                        title="button2 title"
```
CarloPalinckx:
```suggestion
                        title="button title"
```

src/components/Button/index.tsx
```diff
@@ -12,85 +12,20 @@ type PropsType = {
...
-                </Box>
-            )}
-    </>
-);
-
-const Button: SFC<PropsType> = (props): JSX.Element => {
-    const isLink = props.href !== undefined;
+class Button extends Component<PropsType> {
+    public static Cta: typeof StyledCta = StyledCta;
+    public static Plain: typeof StyledPlain = StyledPlain;
```
CarloPalinckx: Would be nice if you could change Plain to Flat here since that aligns with the current prop implementation. No need to change the naming on that now tbh.

PR #225, opened 2018-10-18
Slide in animation

nicokoelewijn:
### This PR:

Adds a new transition: `slide` which animates down from above 

proposed release: `minor`

**Checklist** 🛡

**Adds** ✨
- `slide` animation

**Other changes** 🌀
- button no longer has a scale animation on hover
- removed unused css rules

### DIFF
--- a/src/components/TransitionAnimation/animations/slideDown.ts
+++ b/src/components/TransitionAnimation/animations/slideDown.ts
@@ -10,7 +10,7 @@ const slide = (state: AnimationStateType): string => {
                 }
                 100% {
                     visibility: hidden;
-                    transform: translate3d(0, -150%, 0);
+                    transform: translate3d(0, -100%, 0);
                 }
             }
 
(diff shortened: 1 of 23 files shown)


CarloPalinckx:
Could you add different direction slideIns as well?

CarloPalinckx:
Could you also remove the opacity animation? Maybe with could at some point create different animations that do both, but for now we need slide without opacity 🙂

src/components/TransitionAnimation/animations/slideDown.ts
```diff
@@ -0,0 +1,39 @@
...
+    switch (state) {
+        case 'exiting':
+            return `
+            @keyframes slideOut {
+                0% {
+                    transform: translate3d(0, 0, 0);
+                }
+                100% {
+                    visibility: hidden;
+                    transform: translate3d(0, -150%, 0);
```
CarloPalinckx: Could you revert the transforms to 100%? 150% is not going to work in all cases.

src/components/TransitionAnimation/story.tsx
```diff
@@ -20,7 +20,13 @@ storiesOf('TransitionAnimation', module).add('Default', () => {
         <Box direction="column" alignItems="center">
```
CarloPalinckx: removing alignItems makes for a better story imo :)

```suggestion
         <Box direction="column">
```

PR #226, opened 2018-10-18
Test coverage - 100%

nicokoelewijn:
### This PR:

The tests have once more a 100% test coverage, yay!

proposed release: [`patch`]

**Checklist** 🛡

**Adds** ✨
- Tests for Popover component
- Tests for Select component
- Test for Option component


src/components/Popover/test.tsx
```diff
@@ -106,6 +106,24 @@ describe('Popover', () => {
...
+            <Popover isOpen={false} distance={6} renderContent={(): string => 'Mock content'} />,
+        );
+
+        callbackMap.mousedown({
+            target: document.createElement('div'),
+        });
+
+        component.update();
+
+        expect(component.state('isOpen')).toBe(false);
```
CarloPalinckx: You should try to avoid testing internal behaviour of a component, you should look at unit tests as follows:

### Unit testing, a story of printers, gnomes, and fairies

Imagine you have a printer, you could write a test to assure that:
`given one piece of paper and one jpg of a cat, it produces one printed printed image cat`
This is the only thing you want to test because it doesn't really matter how the printer does it, as long as it provides you with a printed image of cat, it still works.

A test for that could look like:
```js
it('should print a cat image given a piece of paper and a cat image', () => {
    const printer = new Printer();
    const paper = new Paper();
    const catImageUrl = 'https://www.fabickcat.com/content/uploads/2014/10/new-equip-header.jpg';

    const printed = print(paper, catImageUrl);

    expect(printed.cat.length).toEqual(1);
});
```

If a year later, you go ahead and replace al your printer's machinery with an army of gnomes with pencils this test would still pass because it would still produce an image of a cat on paper. Which is great 🎉! You can still promise your customers that their cat images will be successfully delivered to them. No tests have to be changes.

Another year later, a new competitor company enters the market but somehow the cat images they produce are way fluffier. Your beloved customer politely ask you if they could also have fluffier cats, since fluffPrintify(TM) has these as well. 

Your printing function now has extra requirements. You decide that you want to produce at least 75% fluffiness. You write a new test:
```js
it('should print an image with a fluffiness higher than 75%', () => {
    const printer = new Printer();
    const paper = new Paper();
    const catImageUrl = 'https://www.fabickcat.com/content/uploads/2014/10/new-equip-header.jpg';

    const printed = print(paper, catImageUrl);

    expect(printed.fluffiness). toBeGreaterThan(75);
});
```

You find out the harsh truth that gnomes simply cannot produce this amount of fluffiness, the obvious choice is to refactor the printer to use fairies ("duh, you should have gone for fairies in the first place..."). You do the refactor and run your tests... ⌛️ Hooray! both tests pass. You can now tell your customers the good news and sleep easily knowing your product still works. 

In your case:
```js
component.state('isOpen')` 
```
Would equal something like:
```js
it('should print a cat image given a piece of paper and a cat image', () => {
    const printer = new Printer();
    const paper = new Paper();
    const catImageUrl = 'https://www.fabickcat.com/content/uploads/2014/10/new-equip-header.jpg';

    const printed = print(paper, catImageUrl);

    expect(printer.gnomeArmy.length).toBeGreaterThan(200);
});
```
If you would have these kinds of tests. You would have to change the test every time you refactor your printer (I heard orgeJS is the new best thing...). Even though your printer does exactly what you want it to do.

long story short: It doesn't matter if your printer uses gnomes, fairies or ogres. As long as it produces fluffy images of cats 🐱.


**_the end._**

.
. 
. 
. 
. 

That being said: [Here]() are fifty articles on why gnomes and fairies suck and why you should definitely refactor to orgeJS.

PR #227, opened 2018-10-18
resolved errors - work in progress

nicokoelewijn:
### This PR:

Styled Components v4 🎉

**Checklist** 🛡

**Adds** ✨
- Styled Components v4


src/components/Popover/index.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { Manager, Popper, PopperChildrenProps, Reference, ReferenceChildrenProps } from 'react-popper';
+import { Manager, Popper, PopperChildrenProps, Reference, ReferenceChildrenProps, RefHandler } from 'react-popper';
```
hborghols: RefHandler is unused. I assume you imported it to be able to type the refs. It would be cool if those don't have to be any's but I just briefly tried to type them differently otherwise and couldn't figure it out either. Maybe someone else has an idea for this :).

src/components/Popover/index.tsx
```diff
@@ -110,7 +110,10 @@ class Popover extends Component<PropsType, StateType> {
                                     if (ref) this.anchorRef = ref;
                                 }}
                             >
-                                <PopoverAnchor innerRef={ref} stretch={this.props.stretch}>
+                                <PopoverAnchor
+                                    ref={ref as string & ((instance: HTMLDivElement) => undefined)}
```
hborghols: Ah yeah this works. Not sure if I actually prefer this over any but it would be nice to do it consistently, so change all any's to this.

src/components/Popover/index.tsx
CarloPalinckx: Im not sure if callback refs are still supported, you might try using `createRef` here: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

PR #236, opened 2018-10-25
Skeleton Rect

nicokoelewijn:
### This PR:

Adds a new skeleton placeholder. This more general skeleton can be used instead of an image. It's a rectangle, that's it. 

proposed release: `patch`

**Checklist** 🛡

**Adds** ✨
- `<Skeleton.Rect />`


CarloPalinckx:
This pr seems to have commits from a different branch than master. Did you mean to merge this to `1.x`? In the current setup you could merge this to master. However I suggest we remove `Skeleton.Button` with the PR as well since they serve the same purpose. This would mean a merge to `1.x`.

src/components/Skeleton/Rect/style.tsx
```diff
@@ -0,0 +1,74 @@
...
+    right?: string;
+    bottom?: string;
+    margin?: TrblType;
+    borderRadius?: string;
+    height?: PropsType['height'];
+    width?: PropsType['width'];
+    maxHeight?: PropsType['height'];
+    minHeight?: PropsType['minHeight'];
+    maxWidth?: PropsType['maxWidth'];
+    minWidth?: PropsType['minWidth'];
```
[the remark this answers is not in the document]
hborghols: Would it be possible anyway to use a `Box` for his, or a `styled(Box)`? That way you can use a lot of the positioning and sizing logic of Box instead of essentially doing the same here. You can take a look at how the `MultiButton` uses the `Button` under the hood to see what I mean.

src/themes/MerchantTheme/MerchantTheme.test.tsx
```diff
@@ -7,32 +7,21 @@ import { PropsType } from './MerchantTheme.template';
...
-                    hierarchy4: HeadingHierarchyThemeType,
-                    hierarchy5: HeadingHierarchyThemeType,
-                    hierarchy6: HeadingHierarchyThemeType,
-                },
+                hierarchy1: HeadingHierarchyThemeType,
+                hierarchy2: HeadingHierarchyThemeType,
+                hierarchy3: HeadingHierarchyThemeType,
+                hierarchy4: HeadingHierarchyThemeType,
+                hierarchy5: HeadingHierarchyThemeType,
+                hierarchy6: HeadingHierarchyThemeType,
```
hborghols: Why is this in this PR?

src/themes/MerchantTheme/__snapshots__/MerchantTheme.test.tsx.snap
```diff
@@ -5,63 +5,47 @@ exports[`MerchantTheme should render the merchant theme 1`] = `
...
+          "fontSize": "mockValue",
+          "fontWeight": "mockValue",
+          "lineHeight": "mockValue",
+        },
+        "hierarchy6": Object {
+          "color": "mockValue",
+          "fontFamily": "mockValue",
+          "fontSize": "mockValue",
+          "fontWeight": "mockValue",
+          "lineHeight": "mockValue",
```
hborghols: Also not sure why this is here

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -396,63 +402,47 @@ const theme: ThemeType = {
...
+            fontWeight: fontWeight.bold,
+            lineHeight: '21px',
+            color: grey.base,
+        },
+        hierarchy6: {
+            fontFamily: bodyFont,
+            fontSize: '15px',
+            fontWeight: fontWeight.regular,
+            lineHeight: '18px',
+            color: grey.base,
```
hborghols: Or this

nicokoelewijn:
After a fierce discussion, we decided we don't want to inherit from the Box component. The Skeleton should have 100% width and height by default. It should als have width and height props. When a special use-case occurs, the Skeleton can be wrapped in a Box. This way the Skeleton stays a small component without the overhead we would have if we where to inherit from the Box.

hborghols:
Something amazing happens in firefox when you go to the Skeleton.Text story, then the Skeleton.Rect story and then back to the Skeleton.Text story:

screenshot 2018-10-31 at 16 15 20

hborghols:
Something amazing happens in firefox when you visit the Text story, then the Rect story and then the Text story again:

floating-skeleton

hborghols:
When you look at the Text story after looking at the Text story the Text skeletons fly through the screen. I would be curious what causes this and if I have some time to spare I'll investigate.

src/components/Skeleton/story.tsx
```diff
@@ -1,15 +1,16 @@
...
 import Text from '../Text';
 import Button from '../Button';
-
 type DemoPropsType = {
-    loading: boolean;
+    loading?: boolean;
     element: string;
 };
 
+const positionOptions = ['absolute', 'fixed', 'relative', 'static'];
```
hborghols: This is unused

src/components/Skeleton/story.tsx
```diff
@@ -23,14 +24,17 @@ const Demo: SFC<DemoPropsType> = ({ element, loading }): JSX.Element => {
...
         );
-    } else {
+    } else if (element === 'Button') {
         if (loading) {
             return <Skeleton.Button width={number('width', 139)} />;
         }
 
         return <Button title="The button has loaded" variant="primary" />;
+    } else {
+        return <Skeleton.Rect width={text('width', '800px')} height={text('height', '600px')} />;
```
hborghols: It would be nice to also add the loading button to this story

src/components/Skeleton/Rect/style.tsx
```diff
@@ -0,0 +1,45 @@
+import _R from 'react';
+import { StyledComponentClass as _S } from 'styled-components';
+import _T from '../../../types/ThemeType';
+import styled, { withProps } from '../../../utility/styled';
+
+type RectPropsType = {
+    width?: string;
+    height?: string;
```
hborghols: Much cleaner!

hborghols:
> > After a fierce discussion, we decided we don't want to inherit from the Box component. The Skeleton should have 100% width and height by default. It should also have width and height props. When a special use-case occurs, the Skeleton can be wrapped in a Box. This way the Skeleton stays a small component without the overhead we would have if we where to inherit from the Box.
> 
> Alright, seems like you thought about it thoroughly so I support it 👍

I think it turned out very nicely like this :)

CarloPalinckx:
It seems that the keyframes are supposed to be used with a keyframes helper to avoid name collisions (which introduces the funny animation bug).

```
CSS animations with @keyframes aren't scoped to a single component but you still don't want them to be global to avoid name collisions. This is why we export a keyframes helper which will generate a unique instance that you can use throughout your app:
```

Find out more on: https://www.styled-components.com/docs/basics#animations

src/components/TransitionAnimation/test.tsx
```diff
@@ -8,72 +8,73 @@ import slideUp from './animations/slideUp';
...
+        const componentIn = mount(
             <TransitionAnimation show animation={'fade'}>
-                <div>children</div>
+                <div>child</div>
             </TransitionAnimation>,
         );
 
         /* tslint:disable */
-        (expect(componentIn.dive()) as any).toHaveStyleRule('animation', 'fadeIn .3s both');
+        (expect(componentIn.find(StyledAnimation)) as any).toHaveStyleRule('animation', 'iOVxKt .3s both');
```
hborghols: Is this sustainable: are these names guaranteed to stay the same? It seems like it would be wiser not to use the styled-components name in a test. Is it possible to import or mock the rules instead?

nicokoelewijn: I have tried to break this but it seems to stay the same.

CarloPalinckx: It is a hash of the actual content in the style thats applied, so this would break on every change 🙅‍♀️

nicokoelewijn: Haven't found a solution for this yet. This will be solved in another branch. Left it out for now.

package.json
```diff
@@ -67,6 +67,7 @@
     "rimraf": "^2.6.2",
     "source-map-loader": "^0.2.3",
     "styled-components": "^3.2.3",
+    "styled-components-test-utils": "^0.4.1",
```
hborghols: I don't think this is used now right?

PR #248, opened 2018-10-31
Added automatic deployments to now.sh for each push to the repo

hborghols:
### This PR:

Adds automatic deployments of bricks to now.sh on every push to the github repo and included as link in the github PR.

proposed release: `patch`

**Checklist** 🛡

**Adds** ✨
- Automatic deployments of PRs 🌈


src/utility/SubscriptionContext/index.tsx
```diff
@@ -1,91 +1,91 @@
...
-            remove: this.remove,
-            getPayload: this.getPayload,
-        };
-
-        return <Provider value={contextValue}>{this.props.children}</Provider>;
-    }
-}
-
-export { SubscriptionConsumer, SubscriptionProvider, PropsType, StateType, ContextValue, PayloadType, Consumer };
+// import React, { Component, createContext, Consumer } from 'react';
```
CarloPalinckx: uhhhh 😅

hborghols: 😬 Right, this caused some weird TypeErrors during deploying so I commented it out but I forgot all about it, was supposed to be temporary. Will check if it works with the current setup.

hborghols: The TypeError I got was that `createContext` required 1-2 arguments during the initialization of Consumer and Provider. I solved this for now by passing it `undefined` as 'default value'. As far as I could tell from the examples online this shouldn't cause any problems but as `SubscriptionContext` is not used anywhere and as there are no tests for it I'm not a 100% sure if this indeed works..

Issue #250, opened 2018-11-01
TypeScript 3


hborghols:
### Is your feature request related to a problem? Please describe ⚡️
We are still on TypeScript 2.8 but the current version is 3.1 😨. We should think about upgrading some time because **newer === better**. 

### Describe the solution you'd like ✨
New TypeScript

CarloPalinckx:
👍 I've picked up https://github.com/MyOnlineStore/bricks/pull/163 a bit again since sb 4.0. Not much of a hassle to pick this up in there as well.

hborghols:
Cool 😎

hborghols:
This has been added with #251.

PR #251, opened 2018-11-02
Added rgba utility function

### LINKED ISSUE #213: RGB(A) instead of HEX
I'd like to suggest changing the hex codes now used in the theme to be converted to RGB values.
The benefit would be that the common .4 alpha for box-shadows can then become a themeable value.

The result would be something like this: 
```
TextField: {
  focus: {
    boxShadow: '0 0 0 4px rgba(green.lighter1, alpha.boxshadow)',
  }
},
```

We could keep the settings as they are right now, but i do think the chance of error will be reduced.

hborghols:
### This PR:

resolves #213 #250 

proposed release: `minor`

**Checklist** 🛡

**Adds** ✨
- RGBA utility function that converts hex (regular and shorthand), rgb channel strings or rgba channel strings into a valid css rgba string.
- Typescript 3


hborghols:
Yep looks good!

src/utility/rgba/index.ts
```diff
@@ -0,0 +1,32 @@
+type ArgumentsType = [string] | [string, number] | [number, number, number] | [number, number, number, number];
+
+export const hexToRGB = (hex: string): string => {
```
CarloPalinckx: The convention in our codebase is that we camelcase "all the things" 😄. So it would be hexToRgb in this case.

src/utility/rgba/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    const shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
+    hex = hex
```
CarloPalinckx: Try avoiding overwriting function arguments. Some arguments are passed by reference so this could have some side-effects.

src/utility/rgba/index.ts
```diff
@@ -0,0 +1,32 @@
...
+        .replace(/[^0-9A-F]/gi, '')
+        .replace(shorthandRegex, (m: string, r: string, g: string, b: string): string => {
+            return r + r + g + g + b + b;
+        });
```
CarloPalinckx: nit: whitespace

src/utility/rgba/index.ts
hborghols: Oh I thought (for some reason) we did acronyms all upper case or all lower case but 👍

src/utility/rgba/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    const bigint = parseInt(hex, 16);
+    const r = (bigint >> 16) & 255;
+    const g = (bigint >> 8) & 255;
+    const b = bigint & 255;
+
+    return `${r},${g},${b}`;
+};
+
+const rgba = (...args: ArgumentsType): string => {
```
CarloPalinckx: const rgba = (args: ArgumentsType): string => {

hborghols:
Iceboxed until we've upgraded to TypeScript 3 and can use tuples in combination with spreading arguments

PR #254, opened 2018-11-07
Revamped the Table API and added (uncontrolled) sorting

CarloPalinckx:
### This PR:

This PR solves some major problems in the API of the `Table` component. The major problem lied in the definition of the `rows` prop. This allowed you to add JSX elements directly as cell (which seemed like a benefit at first). However this makes most of the data unparseable and therefore unsortable (searchable etc.). This pr solves this by introducing a `columns` prop which contains a map of the properties defined in your rows. IE:

```ts
<Table
    columns={{
        id: { header: 'Product ID' },
        name: { header: 'Name' },
        price: { header: 'Price' },
    }}
    rows={[
        { id: '61651320', price: 19.12, name: 'foo0', image: 'imageurl' },
        { id: '61651321', price: 19.2, name: 'foo1', image: 'imageurl' },
        { id: '61651322', price: 21.12, name: 'foo2', image: 'imageurl' },
        { id: '61651323', price: 22.12, name: 'foo3', image: 'imageurl' },
    ]}
/>,
```

every column in the `columns` prop can now also contain a `sort` property which lets you define a sorting method for that column (also enables the sorting for that column). The custom rendering has been moved to the `render` property in the column as well.

🔥 The column can also contain an `order` property which should be set if you care about the sorting of the columns in the `Table` 

proposed release: `major`

**Checklist** 🛡
- [ ] A designer has seen and approved my changes (check if not applicable)

**Adds** ✨
- Sorting by column

**Changes** 🌀
- `selectable` and `draggable` are removed as prop from the `Table` this is now determined by setting an `onDragEnd` and `onSelection`.
- `onDragEnd` and `onSelection` now receive a mapped rows as first argument of the callback. This allows you to easily sync the state by just doing:

`onDragEnd={(rows) => this.setState({ rows })}`

or:

`onSelection={(rows) => this.setState({ rows })}`

**Removes** 👋
- `headers` and `alignments` prop in favor of two extra properties on the column; `header` and `align`

**Breaking changes** 🔥
- See complete description ⤴️


src/components/Table/Row/index.tsx
```diff
@@ -7,19 +7,18 @@ import StyledRow from './style';
...
 type PropsType = {
-    alignments: Array<'left' | 'center' | 'right'>;
-    cells: Array<ReactNode>;
+    // tslint:disable-next-line
+    columns: { [key: string]: ColumnType<string | number | boolean | undefined> };
+    row: BaseRowType;
     draggable: boolean;
-    checked: boolean;
+    selected: boolean;
     selectable: boolean;
```
nicokoelewijn: Could these props be optional and default to false? In usage this would really be a drag.

src/components/Table/story.tsx
```diff
@@ -1,91 +1,119 @@
...
 };
 
 type PropsType = {
     draggable: boolean;
     selectable: boolean;
+    sortable: boolean;
+    custom: boolean;
 };
 
-const actions = (
```
nicokoelewijn: To bad the buttons have been removed in the story. Any example of the buttons in the table would be great imo.

src/components/Table/story.tsx
```diff
@@ -1,91 +1,119 @@
...
-                onSelection={(rows): void => this.setState({ rows })}
+                onDragEnd={this.props.draggable ? (rows): void => this.setState({ rows }) : undefined}
+                onSelection={this.props.selectable ? (rows): void => this.setState({ rows }) : undefined}
             />
         );
     }
 }
 
-storiesOf('Table', module).add('Default', () => {
-    return <Demo selectable={boolean('selectable', false)} draggable={boolean('draggable', true)} />;
```
nicokoelewijn: There are a few impossible combinations if all the knobs where to be enabled, but an i/o knob per functionality would be great. Just to see nothing breaks if the functionality where to be toggled.

src/components/Table/Row/index.tsx
CarloPalinckx: These are only for internal usage in the Table 🙂.  Making these required saves a few if statements in the actual code which makes it more readable imo.

src/components/Table/story.tsx
CarloPalinckx: I can add them to the custom rendering story 👍

src/components/Table/Row/index.tsx
nicokoelewijn: Oh, my bad. carry on then 🙂

src/components/Table/Headers/index.tsx
```diff
@@ -0,0 +1,175 @@
...
+                            if (
+                                this.props.columns[a].order === undefined ||
+                                this.props.columns[b].order === undefined
+                            ) {
+                                return -1;
+                            }
+
+                            return (this.props.columns[a].order as number) - (this.props.columns[b].order as number);
+                        })
+                        .map(
```
nicokoelewijn: The indenting from here becomes rather deep. Perhaps you could refactor this code a bit?

digibeuk: there  is no such thing as `a little bit of refactoring` :stuck_out_tongue:

PR #257, opened 2018-11-14
removed onFocus- onBlurCapture from wrapper to prevent double focus

### LINKED ISSUE #256: Multiple onBlur & onFocus calls on TextField
### Bricks version 📦
0.6.2

### Describe the bug 🔍
The `StyledWrapper` captures the blur and focus events and executes the onBlur and onFocus props, while these are already called on the input. So it results in multiple calls. The onFocus is even called a third time due to the onClickCapture.

The good thing about the click capture is that even when you click on a prefix or suffix, the field is focussed. The rest seems unnecessary.

### To Reproduce 💥
1. Go to story of TextField and add a onBlur and onFocus with a console log or something
2. Focus and blur and behold the console.

### Expected behavior ✨
I would expect it to call the onBlur and onFocus once per event. But I would also expect the field to focus when clicking on a prefix. So I would like to suggest to remove the `onBlurCapture` and `onFocusCapture`, but keep the `onClickCapture` to force a focus on the input.

nicokoelewijn:
### This PR:

removed onFocus- onBlurCapture from wrapper to prevent double focus.
resolves #256  

proposed release: `patch`

**Checklist** 🛡

**Changes** 🌀
- removed onFocus- onBlurCapture from wrapper to prevent double focus

### DIFF
--- a/src/components/TextField/index.tsx
+++ b/src/components/TextField/index.tsx
@@ -7,7 +7,6 @@ import { StyledInput, StyledWrapper, StyledAffix, StyledAffixWrapper } from './s
 import withCurrencyFormatting, { WithCurrencyFormattingType } from './formatters/withCurrencyFormatting';
 import withNumberFormatting, { WithNumberFormattingType } from './formatters/withNumberFormatting';
 import Icon from '../Icon';
-import BreakpointProvider from '../BreakpointProvider';
 
 type PropsType = {
     value: string;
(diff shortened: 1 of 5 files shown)


src/components/TextField/index.tsx
```diff
@@ -64,12 +63,9 @@ class TextField extends Component<PropsType, StateType> {
                     focus={this.state.focus}
                     disabled={this.props.disabled}
                     severity={this.props.feedback ? this.props.feedback.severity : 'success'}
-                    onFocusCapture={this.handleFocus}
-                    onBlurCapture={this.handleBlur}
-                    onClick={this.handleFocus}
                 >
                     {this.props.prefix && (
-                        <StyledAffixWrapper disabled={this.props.disabled}>
+                        <StyledAffixWrapper onClick={this.handleFocus} disabled={this.props.disabled}>
```
woutsluiter: The onFocus is still called twice when clicking on a prefix. I think you need to move the `this.inputRef.focus()` in `handleFocus` to a separate `forceFocus` and pass that here.

src/components/TextField/test.tsx
```diff
@@ -36,6 +36,14 @@ describe('TextField', () => {
         expect(component.find(StyledWrapper).prop('focus')).toBe(true);
     });
 
+    it('should render an active state when focussed by clicking an affix', () => {
```
woutsluiter: nit: I think this should be 'focus'

PR #262, opened 2018-11-20
Added auto dismiss prop to Toaster

### LINKED ISSUE #259: Toaster auto dismissal
It would be nice if notifications where to dismiss after the message. This would be an additional prop for the `Toaster`. 
This feature is inspired by the Hubspot alert component (https://canvas.hubspot.com/components/alerts-messaging#floating).
By making this an optional prop, the Toaster will dismiss automatically after x seconds, that way less importat notifications won't have to be closed.  

Add a prop `autoDismiss`. When the time has passed the Toaster would become invisible again. 

A different solution would be to create a `dismissTime` prop that would set the time in seconds before the toaster will dismiss. 
I do think that it would be best if the interval  is the same everywhere. 

### Additional context 🖼
Example of how this would work: https://canvas.hubspot.com/components/alerts-messaging#floating, _click the show default notification_

nicokoelewijn:
### This PR:

resolves #259

proposed release: [`patch`]

**Checklist** 🛡

**Adds** ✨
- Auto dismiss prop

### DIFF
--- a/src/components/Toaster/index.tsx
+++ b/src/components/Toaster/index.tsx
@@ -42,7 +42,7 @@ class Toaster extends Component<PropsType> {
     };
 
     public componentDidMount = (): void => {
-        if (this.props.autoDismiss === true) setTimeout((): void => this.closeAction(), 6000);
+        if (this.props.autoDismiss) setTimeout((): void => this.closeAction(), 6000);
     };
 
     public render(): JSX.Element {
--- a/src/components/Toaster/index.tsx
+++ b/src/components/Toaster/index.tsx
@@ -63,13 +63,14 @@ class Toaster extends Component<PropsType> {
                                         </Box>
                                     )}
                                     <Box
+                                        style={{ display: breakpoint === 'small' ? 'block' : '' }}
                                         direction={breakpoint === 'small' ? 'column' : 'row'}
                                         justifyContent="center"
                                         alignContent="center"
                                     >
                                         <Box
-                                            direction="column"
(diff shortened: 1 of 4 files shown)


src/components/Toaster/index.tsx
```diff
@@ -17,83 +17,93 @@ type PropsType = {
...
-                                    justifyContent="center"
-                                    alignContent="center"
-                                >
-                                    <Box direction="column" margin={breakpoint === 'small' ? trbl(12) : trbl(18, 12)}>
-                                        <Text strong>{props.title}</Text>
-                                        <Text>{props.message}</Text>
-                                    </Box>
-                                    {props.buttonTitle && (
+    public componentDidMount = (): void => {
+        if (this.props.autoDismiss === true) setTimeout((): void => this.closeAction(), 6000);
```
hborghols:
```suggestion
        if (this.props.autoDismiss) setTimeout((): void => this.closeAction(), 6000);
```

src/components/Toaster/test.tsx
```diff
@@ -18,6 +20,26 @@ describe('Toaster', () => {
         expect(clickMock).toHaveBeenCalled();
     });
 
+    it('should be possible to automatically dismiss', () => {
```
hborghols:
```suggestion
    it('should automatically close when autoDismiss is true', () => {
```

src/components/Toaster/index.tsx
nicokoelewijn: without `=== true`, it would only check if the prop is set. So, it would also trigger on `autoDismiss={false}`

PR #265, opened 2018-11-26
Added Seperated component

### LINKED ISSUE #261: Turn RadioButtonGroup into a general Separated component
We are currently starting to see patterns in grouping behavior. We support the `RadioButonGroup` at the moment which actually has nothing to do with the `RadioButton` apart from naming 😄  it just maps children in a `Box` and applies padding.

I suggest we turn the `RadioButtonGroup` into a component named `Separated`. This component could provide a general way of visually "grouping" any element. The current RadioButtonGroup only supports margin as saperator. We could later add a visual divider to this. This could also make the MessageStream obsolete (which is nice) and it could possibly solve some issues we are having when using a Checkbox(group) in a FormRow. 

We could optionally allow for a prop to be set that enables/disables the leading separator (be it margin or a divider). We could also look into requiring children to be added (If that is even possible in a sane way). This would prevent people from abusing it to position things instead of using a `Box`

```ts
type Props = {
    by?: 'margin'|'divider'; // divider could be added later
    before?: boolean;
    after?: boolean;
    children: ReactElement
}
```

Would like some ideas for the prop name. I'm not really sure if it the proper name for what it does.

We could alternatively double export RadioButtonGroup as CheckBoxGroup

CarloPalinckx:
### This PR:

resolves #261 
proposed release: `minor`

**Checklist** 🛡

**Adds** ✨
- A `Seperated` component according to discussion #261

**Changes** 🌀
- Marks `RadioButtonGroup` as deprecated in storybook.

### DIFF
--- a/src/components/Separated/index.tsx
+++ b/src/components/Separated/index.tsx
@@ -1,4 +1,4 @@
-import React, { SFC, Children } from 'react';
+import React, { FunctionComponent, Children } from 'react';
 import Box from '../Box';
 import trbl, { TrblType } from '../../utility/trbl';
 
@@ -19,7 +19,7 @@ const calculateMargin = (count: number, index: number, before?: boolean, after?:
     return trbl(margin, 0, 0, 0);
 };
 
-const Separated: SFC<PropsType> = ({ children, before, after }): JSX.Element => {
+const Separated: FunctionComponent<PropsType> = ({ children, before, after }): JSX.Element => {
     const count = Children.count(children);
 
     return (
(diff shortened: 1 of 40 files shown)


hborghols:
Maybe you could add a boundary around the items, similar to the ScrollBox or ButtonGroup stories. That way you can make `after` visible when you enable it.

src/components/Separated/index.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { SFC, Children } from 'react';
```
hborghols: SFC is deprecated in favor of FunctionComponent

PR #266, opened 2018-11-27
Fixed TextField.Number to allow it to be empty

hborghols:
### This PR:

Fixes a behavior in `TextField.Number` that didn't allow it to be empty. It also saves the last valid input and restores that on blur when invalid input is given (as discussed with luuk).

proposed release: `patch`

**Checklist** 🛡

**Changes** 🌀
- Fixes `Textfield.Number` behavior.

### DIFF
--- a/src/components/TextField/formatters/withNumberFormatting/index.tsx
+++ b/src/components/TextField/formatters/withNumberFormatting/index.tsx
@@ -11,8 +11,7 @@ type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, Omit
 
 type StateType = {
     value: string;
-    savedValue: string;
-    inputLength: number;
+    savedValue: number;
 };
 
 type WithNumberFormattingType = ComponentClass<PropsType>;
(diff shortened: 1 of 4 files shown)


src/components/TextField/formatters/withNumberFormatting/index.tsx
```diff
@@ -22,20 +24,27 @@ const withNumberFormatting = (Wrapped: ComponentType<TextFieldPropsType>): Compo
...
                 value: `${nextProps.value}`,
             };
         }
 
         private handleChange = (value: string): void => {
             const parsedValue = parseInt(value, 10);
-
             if (isNaN(parsedValue)) {
                 this.setState({ value: '' });
+                this.props.onChange(NaN);
```
CarloPalinckx: This will probably break stuff. Isn't there something valid we can return here? Maybe call it with the previously entered valid value?

src/components/TextField/formatters/withNumberFormatting/index.tsx
```diff
@@ -11,6 +11,8 @@ type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, Omit
 
 type StateType = {
     value: string;
+    savedValue: string;
+    inputLength: number;
```
woutsluiter: This state is not used anywhere

src/components/TextField/formatters/withNumberFormatting/index.tsx
hborghols: Yep that works too

PR #273, opened 2018-12-03
Removed checkbox class

### LINKED ISSUE #272: Checkbox offset
### Bricks version 📦
0.6.3

### Describe the bug 🔍
Checkboxes in the FormRow have an offset. This is caused by the checkbox class. 
This  can be solved by implementing the Separate component and removing the checkbox class.

### To Reproduce 💥
Check the offset of the checkbox in a FormRow

### Expected behavior ✨
there shouldn't be an offset on the checkbox. 

### Screenshots 📸
screenshot 2018-12-03 at 15 59 59

### Context👨‍💻
 - Every context

nicokoelewijn:
### This PR:

resolves #272 

proposed release: `patch`

**Checklist** 🛡

**Changes** 🌀
- Implements Separated component in FormRow

**Removes** 👋
- Depracted `.checkbox` class in FormRow and therefore it's css

**Breaking** 🔥
- Removed .checkbox classname and styling

### DIFF
--- a/src/components/FormRow/story.tsx
+++ b/src/components/FormRow/story.tsx
@@ -269,7 +269,7 @@ class DemoComponent extends Component<PropsType, StateType> {
                             </label>
                         }
                         field={
-                            <Separated before>
+                            <Box margin={trbl(0, 12, 0, 0)}>
                                 <Checkbox
                                     onChange={(): void =>
                                         this.setState({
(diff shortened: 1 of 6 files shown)


hborghols:
The Separate component is intended for grouping things so the way you used it here is not it's use case. You can just use a regular old Box to give the checkbox it's top margin. Also: I would see this as a breaking change because it breaks the positioning of all checkboxes in forms currently in use.

src/components/FormRow/story.tsx
```diff
@@ -9,6 +9,7 @@ import Box from '../Box';
 import TextField from '../TextField';
 import Toggle from '../Toggle';
 import trbl from '../../utility/trbl';
+import Separated from '../Separated';
```
hborghols: You don't need this anymore

PR #276, opened 2018-12-05
Ship separate ES5/ES2015+ bundles

hborghols:
### This PR:

Adds support to ship separate ES5 (as `[name]index.js`) and ES2015+ (`as [name]modern.js`) bundles. This allows us to only ship ES2015+ code to users with modern browsers and thus reduce the required data. This PR adds build, compile and bundle scripts in the `scripts` folder. The bundle script contains the webpack configs for the respective bundles.  

proposed release: `major`

**Checklist** 🛡
- [ ] I have tested my addition in all supported browsers and for responsiveness (Chrome, Firefox, Safari, Edge, IE11 and mobile browsers)

**Adds** ✨
- Support for shipping modern and legacy JS bundles.

**Changes** 🌀
- The build config

**Removes** 👋
- `webpack/webpack.config.js` and replaces it with `scripts/bundles.js`.


scripts/bundles.js
```diff
@@ -0,0 +1,210 @@
...
+        new webpack.DefinePlugin({
+            'process.env.NODE_ENV': JSON.stringify('production'),
+        }),
+        new PeerDepsExternalsPlugin(),
+    ],
+};
+
+const modernConfig = Object.assign({}, baseConfig, {
+    output: {
+        filename: '[name]index.es2015.js',
```
woutsluiter:
```suggestion
        filename: '[name]modern.js',
```

woutsluiter: Gives a nicer import imo: `import Button from '@myonlinestore/bricks/lib/components/Button/modern';`

package.json
```diff
@@ -90,13 +90,13 @@
...
     "test": "jest --config=./config/jest/jest.config.json",
     "test-coverage": "jest --config=./config/jest/jest.config.json --coverage --runInBand --bail",
-    "build": "rimraf ./lib && webpack --config ./config/webpack/webpack.config.js",
     "lint-fix": "tslint --fix --project ./config/typescript/tsconfig.json --config ./config/typescript/tslint.json ./src/**/*.{ts,tsx}",
     "lint": "tslint --project ./config/typescript/tsconfig.json --config ./config/typescript/tslint.json ./src/**/*.{ts,tsx}",
     "watch": "start-storybook -p 9001 -c ./config/storybook",
     "deploy-storybook": "storybook-to-ghpages",
     "build-now": "build-storybook -c ./config/storybook -o /public",
     "build-storybook": "build-storybook -c ./config/storybook -o .out",
+    "build": "rimraf ./lib && node scripts/build.js",
```
woutsluiter: Not specific for this line but could you change the `files` key from `dist` to `lib`?

scripts/bundles.js
```diff
@@ -0,0 +1,210 @@
...
+            sourceMap: true,
+        }),
+        new webpack.DefinePlugin({
+            'process.env.NODE_ENV': JSON.stringify('production'),
+        }),
+        new PeerDepsExternalsPlugin(),
+    ],
+};
+
+const modernConfig = Object.assign({}, baseConfig, {
```
woutsluiter: The config isn't deep merged, so the plugins are overwritten. Maybe you could use https://github.com/survivejs/webpack-merge for this

scripts/bundles.js
```diff
@@ -0,0 +1,210 @@
...
+            analyzerMode: 'static',
+            openAnalyzer: false,
+            reportFilename: '../reports/webpack/statistics-tree-modern.html',
+        }),
+    ],
+});
+
+const legacyConfig = Object.assign({}, baseConfig, {
+    output: {
+        filename: '[name]index.es5.js',
```
woutsluiter: And this could be `index.js`

hborghols:
It seems like the modern and legacy bundles are nearly identical for some reason. This is iceboxed until we've had time for future research.

PR #277, opened 2018-12-11
Created spinner component

### LINKED ISSUE #269: Spinner component
Blocks #270 

We're missing ways to indicate a loading state when saving a form or validating a textfield. Right now a text replacement like "...Almost done" along with a disabled state is used, but this is not ideal. The background switch is too heavy and with long during requests it looks like the application froze.

To indicate loading states with continuous feedback we want to use an animated spinner. Basically we want this: https://evergreen.segment.com/components/spinner. It uses a svg with a circle component which has an css animation on the stroke-dash-offset for the elastic effect. It should also have a `color` prop so it can be used in different contexts.

A progressbar would also be nice, but not suitable for an indeterminate loading state

nicokoelewijn:
### This PR:

resolves #269

proposed release: `patch`

**Checklist** 🛡

**Adds** ✨
- Spinner component

### DIFF
--- a/src/components/Spinner/index.tsx
+++ b/src/components/Spinner/index.tsx
@@ -1,14 +1,15 @@
 import React, { FunctionComponent } from 'react';
-import StyledSpinner, { SpinnerSvg, SpinnerCircle, SpinnerPropsType } from './style';
+import StyledSpinner, { SpinnerSvg, SpinnerCircle } from './style';
+
+type SpinnerPropsType = {
+    color: string;
+};
 
 const Spinner: FunctionComponent<SpinnerPropsType> = (props): JSX.Element => (
     <StyledSpinner aria-busy="true" aria-live="polite">
         <SpinnerSvg
             // @ts-ignore
-            x="0px"
-            y="0px"
             viewBox="0 0 150 150"
-            fill="transparent"
             color={props.color}
         >
             <SpinnerCircle cx="75" cy="75" r="60" />
@@ -17,3 +18,4 @@ const Spinner: FunctionComponent<SpinnerPropsType> = (props): JSX.Element => (
 );
 
 export default Spinner;
+export { SpinnerPropsType };
--- a/src/components/Spinner/index.tsx
+++ b/src/components/Spinner/index.tsx
@@ -1,11 +1,11 @@
 import React, { FunctionComponent } from 'react';
 import StyledSpinner, { SpinnerSvg, SpinnerCircle } from './style';
 
-type SpinnerPropsType = {
+type PropsType = {
     color: string;
 };
 
-const Spinner: FunctionComponent<SpinnerPropsType> = (props): JSX.Element => (
+const Spinner: FunctionComponent<PropsType> = (props): JSX.Element => (
     <StyledSpinner aria-busy="true" aria-live="polite">
         <SpinnerSvg
             // @ts-ignore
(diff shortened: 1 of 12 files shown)


src/components/Spinner/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+const Spinner: FunctionComponent<SpinnerPropsType> = (props): JSX.Element => (
+    <StyledSpinner aria-busy="true" aria-live="polite">
+        <SpinnerSvg
+            // @ts-ignore
+            x="0px"
+            y="0px"
+            viewBox="0 0 150 150"
+            fill="transparent"
+            color={props.color}
```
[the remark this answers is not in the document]
hborghols: Good point, I would verify this with design

src/components/Spinner/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { FunctionComponent } from 'react';
+import StyledSpinner, { SpinnerSvg, SpinnerCircle, SpinnerPropsType } from './style';
+
+const Spinner: FunctionComponent<SpinnerPropsType> = (props): JSX.Element => (
+    <StyledSpinner aria-busy="true" aria-live="polite">
+        <SpinnerSvg
+            // @ts-ignore
+            x="0px"
```
hborghols: Why do you need this (and the `y` below)? It seems to work without them for me.

src/components/Spinner/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+            y="0px"
+            viewBox="0 0 150 150"
+            fill="transparent"
```
hborghols: You also set this in `style.tsx`, I would remove one of the two

src/components/Spinner/index.tsx
```diff
@@ -0,0 +1,19 @@
+import React, { FunctionComponent } from 'react';
+import StyledSpinner, { SpinnerSvg, SpinnerCircle, SpinnerPropsType } from './style';
+
+const Spinner: FunctionComponent<SpinnerPropsType> = (props): JSX.Element => (
```
woutsluiter: I think you forgot a PropsType

src/components/Spinner/story.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { storiesOf } from '@storybook/react';
+import Box from '../Box';
+import React from 'react';
+import Spinner from '.';
+
+storiesOf('Spinner', module).add('Default', () => {
+    const size = text('size', '100px');
+
+    return (
+        <Box width={size} height={size}>
```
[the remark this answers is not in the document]
nicokoelewijn: I made it the same way the skeleton rectangle is made, with only the core functionality and no unnecessary positioning props. But i'll add an indication that it's part of the story.

src/components/Spinner/index.tsx
nicokoelewijn: This is based on the https://evergreen.segment.com/components/spinner/ component, but i can't find any added value either, so i'll remove it.

src/components/Spinner/index.tsx
nicokoelewijn: No i wrote it in the style, but i'll move it to the index.

src/components/Spinner/index.tsx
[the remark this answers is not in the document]
nicokoelewijn: Luuk approves. It should match the font color. So if the component remains as is, the color prop can be passed on to it

src/components/Spinner/index.tsx
hborghols: Can you rename it to `PropsType` to make it consistent with all other components?

PR #279, opened 2018-12-13
Created loading state for Button

### LINKED ISSUE #270: Button loading state
Is blocked by #269 

We're missing ways to indicate a loading state when saving a form or validating a textfield. Right now a text replacement like "...Almost done" along with a disabled state is used, but this is not ideal. The background switch is too heavy and with long during requests it looks like the application froze.

To indicate a loading state on a button we can use the spinner from #269. It should be positioned in dead center without affecting the buttons box-model, so the text should remain in place, just transparent. The spinner color can be derived from the text color.

### Additional context 🖼
Example: https://polaris.shopify.com/components/actions/button#navigation (select loading state)

nicokoelewijn:
### This PR:

resolves #270 

proposed release: `minor`

**Checklist** 🛡

**Adds** ✨
- loading state for `Button`

### DIFF
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -1,12 +1,14 @@
 import React, { Children, FunctionComponent } from 'react';
 import StyledButton, { StyledAnchor } from './style';
 import Icon, { MediumPropsType } from '../Icon';
+import ThemeType from '../../types/ThemeType';
+import { withTheme } from 'styled-components';
 import Box from '../Box';
-import theme from '../../themes/MosTheme/MosTheme.theme';
 import trbl from '../../utility/trbl';
 import Spinner from '../Spinner';
 
 type PropsType = {
+    theme: ThemeType;
     className?: string;
     compact?: boolean;
     loading?: boolean;
@@ -50,6 +52,7 @@ const ButtonContents: FunctionComponent<ButtonContentsType> = (props): JSX.Eleme
 const Button: FunctionComponent<PropsType> = (props): JSX.Element => {
     const isLink = props.href !== undefined;
     const subVariant = props.flat ? 'flat' : 'regular';
+
     const clickAction = (): void => {
         if (props.action !== undefined && props.disabled !== true && props.loading !== true) {
             props.action();
@@ -97,9 +100,6 @@ const Button: FunctionComponent<PropsType> = (props): JSX.Element => {
         >
             {props.loading && (
                 <Box
-                    grow={0}
-                    shrink={0}
-                    basis="35px"
                     justifyContent="center"
                     alignItems="center"
                     position="absolute"
@@ -109,7 +109,7 @@ const Button: FunctionComponent<PropsType> = (props): JSX.Element => {
                     bottom="0"
                     padding={trbl(6)}
                 >
-                    <Spinner color={theme.Button[props.variant][subVariant].idle.color} />
+                    <Spinner color={props.theme.Button[props.variant][subVariant].idle.color} />
                 </Box>
             )}
             <ButtonContents title={props.title} icon={props.icon} iconAlign={props.iconAlign}>
@@ -119,5 +119,5 @@ const Button: FunctionComponent<PropsType> = (props): JSX.Element => {
     );
 };
 
-export default Button;
+export default withTheme(Button);
 export { PropsType };
--- a/src/components/Button/index.tsx
+++ b/src/components/Button/index.tsx
@@ -28,11 +28,29 @@ type PropsType = {
 type ButtonContentsType = {
     icon?: PropsType['icon'];
     iconAlign?: PropsType['iconAlign'];
+    subVariant: 'flat' | 'regular';
+    variant: PropsType['variant'];
+    loading: PropsType['loading'];
+    theme: PropsType['theme'];
     title: PropsType['title'];
 };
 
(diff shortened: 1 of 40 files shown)


src/components/Button/test.tsx
```diff
@@ -14,6 +14,42 @@ describe('Button', () => {
...
+            <Button loading icon={'checkmark'} title="Foo Bar?" action={clickMock} variant="warning">
+                Foo Bar
+            </Button>,
+        );
+        linkComponent.simulate('click');
+        buttonComponent.simulate('click');
+        expect(clickMock).not.toHaveBeenCalled();
+    });
+
+    // it('should update button text when loading starts', () => {
```
hborghols: Why is this test commented out?

hborghols: I guess it's not necessary anymore and can be deleted?

hborghols: This looks like it can be deleted.

src/components/Button/index.tsx
```diff
@@ -81,10 +90,27 @@ const Button: FunctionComponent<PropsType> = (props): JSX.Element => {
             type="button"
             disabled={props.disabled}
             flat={props.flat}
-            color={props.color}
+            color={props.loading ? 'transparent' : props.color}
             icon={props.icon}
             id={props.id}
         >
+            {props.loading && (
+                <Box
```
hborghols: Seeing this Box makes me wonder if it shouldn't be moved inside the Spinner component after all. I don't see that as blocking for this PR but maybe we can discuss it with the rest next week.

src/components/Button/index.tsx
```diff
@@ -2,11 +2,13 @@ import React, { Children, FunctionComponent } from 'react';
 import StyledButton, { StyledAnchor } from './style';
 import Icon, { MediumPropsType } from '../Icon';
 import Box from '../Box';
+import theme from '../../themes/MosTheme/MosTheme.theme';
 import trbl from '../../utility/trbl';
-
+import Spinner from '../Spinner';
 type PropsType = {
```
CarloPalinckx: nit: whitespace

src/components/Button/index.tsx
CarloPalinckx: Why would you want to include this in the Spinner? Seems a bit unconcerning to the Spinner to handle its own positioning since that's what the box is for 🤷‍♂️

src/components/Button/index.tsx
```diff
@@ -81,10 +91,27 @@ const Button: FunctionComponent<PropsType> = (props): JSX.Element => {
...
-            color={props.color}
+            color={props.loading ? 'transparent' : props.color}
             icon={props.icon}
             id={props.id}
         >
+            {props.loading && (
+                <Box
+                    grow={0}
+                    shrink={0}
+                    basis="35px"
```
CarloPalinckx: the first 3 props are redundant since:
```ts
left="0"
top="0"
right="0"
bottom="0"
```
already stretches the Box

src/components/Button/index.tsx
```diff
@@ -81,10 +91,27 @@ const Button: FunctionComponent<PropsType> = (props): JSX.Element => {
...
+                    justifyContent="center"
+                    alignItems="center"
+                    position="absolute"
+                    left="0"
+                    top="0"
+                    right="0"
+                    bottom="0"
+                    padding={trbl(6)}
+                >
+                    <Spinner color={theme.Button[props.variant][subVariant].idle.color} />
```
CarloPalinckx: this is going to break all theme's other than the mos theme, you are importing the MosTheme here.

src/components/Button/index.tsx
```diff
@@ -64,15 +74,18 @@ const Button: FunctionComponent<PropsType> = (props): JSX.Element => {
                 flat={props.flat}
                 id={props.id}
             >
-                <ButtonContents title={props.title} icon={props.icon} iconAlign={props.iconAlign}>
-                    {props.children}
-                </ButtonContents>
+                {(props.loading && <Spinner color={'currentColor'} />) || (
```
CarloPalinckx: `currentColor` won't work since you aren't applying this prop to `color` but to `stroke`, you have to use the theme here as well.

CarloPalinckx: The positioning also seems to be busted for the anchor implementation.

src/components/Button/index.tsx
CarloPalinckx: As in a "centered" | "inline" prop?

hborghols: For example yeah

nicokoelewijn: Will be resolved in another PR. Please open a RFC for this :)

src/components/Button/index.tsx
```diff
@@ -46,24 +51,43 @@ const ButtonContents: FunctionComponent<ButtonContentsType> = (props): JSX.Eleme
...
                 className={props.className}
                 href={props.href}
                 target={props.target}
+                color={props.loading ? 'transparent' : props.color}
                 disabled={props.disabled}
                 flat={props.flat}
                 id={props.id}
             >
+                {props.loading && (
+                    <Box
```
CarloPalinckx: It looks like this is a bit of duplicated code, since this behavior is always the same for the anchor as well as the button you could put it in the `ButtonContents` which possibly avoids differences between the anchor and the button rendering :)

PR #282, opened 2018-12-18
Select handles change events

CarloPalinckx:
### This PR:

Add handling of a (simulated) change event to the Select. This makes it easier to test and possibly makes it more interchangeable with an actual html select

proposed release: `minor`

**Checklist** 🛡

**Adds** ✨
- Change event handling for Select
- A test to cover this

### DIFF
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, KeyboardEvent, Component, RefObject, createRef } from 'react';
+import React, { ChangeEvent, KeyboardEvent, Component, RefObject, createRef, FormEvent } from 'react';
 import { createPortal } from 'react-dom';
 import Box from '../Box';
 import ScrollBox from '../ScrollBox';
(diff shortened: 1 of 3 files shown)


src/components/Select/index.tsx
```diff
@@ -178,6 +183,7 @@ class Select<GenericOptionType extends OptionBaseType> extends Component<PropsTy
                 isDisabled={this.props.disabled}
                 isOpen={this.state.isOpen}
                 onKeyDownCapture={this.handleKeyPress}
+                onChange={this.handleChangeEvent}
```
nicokoelewijn: A knob for this behaviour would be very welcome :)

hborghols: What kind of knob? In the story you mean? As this is a feature mainly meant for testing I don't really see the added value in that.

CarloPalinckx: We talked about a knob that allows you to set the value so you could test if the value is handled properly when changing the prop. However, on second thought it looks like a bit too much work to allow for a decent knob and also keep supporting manual input. @nicokoelewijn if it's okay with you i'd rather look at this some other time because I think there are more cases like this in our current storybook that we could take a look at. Perhaps this would combine well with writing documentation for 1.0.0.

nicokoelewijn: 👍

hborghols: 👍

PR #283, opened 2018-12-18
Removed storyshots

CarloPalinckx:
### This PR:

Removes storyshots 

proposed release: `patch`

**Checklist** 🛡

**Removes** 👋
- storyshots


hborghols:
Should we still adjust the coverage threshold so new PRs don't always fail CI until we've regained coverage with new tests? We could set it to 95% for example.

CarloPalinckx:
The threshold is already 80%, This pr is declined because it is a negative change in the percentage. I guess we could just make an exception for this special case where a negative change is actually a good thing 😄

hborghols:
Haha alright. I'll allow it. If you update the branch I'll merge it 👍

PR #287, opened 2018-12-21
Textfield.Currency supports minor units

robinvdv12:
### This PR:

When dealing with currency values, it is best practice to work with the the amount in minor values. This pr adds a prop that internalizes this behavior in the input.

- Component `TextField.Currency` now also supports an optional `minor` prop, to enable working with minor units.

**Checklist** 🛡

### DIFF
--- a/src/components/TextField/formatters/withCurrencyFormatting/test.tsx
+++ b/src/components/TextField/formatters/withCurrencyFormatting/test.tsx
@@ -250,7 +250,7 @@ describe('withCurrencyFormatting', () => {
         expect(component.find('input').prop('value')).toEqual('-19.12');
     });
 
-    it("should divide the value property by ten to the power of the maximum fraction digits, when the 'minor' property is true", () => {
+    it('should use the value in minor units, when the "minor" prop is set', () => {
         const CurrencyField = withCurrencyFormatting(TextField);
         // Work-around because the Intl polyfil doesn't support resolvedOptions()
         CurrencyField.prototype.setFormatter = () => {
@@ -265,7 +265,6 @@ describe('withCurrencyFormatting', () => {
 
         const component = mountWithTheme(
             <CurrencyField
-                disableNegative={false}
                 name=""
                 value={2554}
                 locale="nl-NL"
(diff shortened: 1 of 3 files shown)


src/components/TextField/formatters/withCurrencyFormatting/test.tsx
```diff
@@ -249,4 +249,40 @@ describe('withCurrencyFormatting', () => {
...
+
+        expect(component.find('input').prop('value')).toEqual('25.54');
+
+        component.find('input').simulate('change', {
+            target: {
+                value: '1908',
+            },
+        });
+
+        expect(component.find('input').prop('value')).toEqual('1908');
```
CarloPalinckx: You might want to add an assertion that expects if changeMock has been called with the value in minor units.

src/components/TextField/formatters/withCurrencyFormatting/test.tsx
```diff
@@ -249,4 +249,40 @@ describe('withCurrencyFormatting', () => {
...
+                    maximumFractionDigits: 2,
+                }),
+            };
+        };
+
+        const changeMock = jest.fn();
+
+        const component = mountWithTheme(
+            <CurrencyField
+                disableNegative={false}
```
CarloPalinckx: You can leave this prop out here, doesn't really add anything :)

src/components/TextField/formatters/withCurrencyFormatting/test.tsx
```diff
@@ -249,4 +249,40 @@ describe('withCurrencyFormatting', () => {
 
         expect(component.find('input').prop('value')).toEqual('-19.12');
     });
+
+    it("should divide the value property by ten to the power of the maximum fraction digits, when the 'minor' property is true", () => {
```
CarloPalinckx: You don't really have to explain how the dividing works here IMO, maybe you improve the description by naming it something along the lines of:
```suggestion
    it('should use the value in minor units, when the "minor" prop is set', () => {
```
property can also be prop to avoid confusion with class properties.

PR #288, opened 2018-12-21
Improved modal responsiveness

Dreamdealer:
### This PR:

- Added support to render an array of buttons
- Added a way to skip the breakpointprovider in buttongroups
- Made the styling conform design specs
- Made the title editable in storybook
- Deprecation warning in console and docs

**Breaking changes** 🔥
- None.

- Component `modal` now also accepts an array of Reactnodes to use in the button bar.

- Conformed the styling to the design specs.

**Checklist** 🛡

### DIFF
--- a/src/components/Modal/index.tsx
+++ b/src/components/Modal/index.tsx
@@ -98,7 +98,7 @@ class Modal extends Component<PropsType> {
                                         {this.props.children}
                                     </Box>
                                 </ScrollBox>
-                                {this.props.renderFixed || this.props.buttons && (
+                                {(this.props.renderFixed || this.props.buttons) && (
                                     <Contrast>
                                         <Box
                                             direction="column"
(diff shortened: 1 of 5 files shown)


src/components/Modal/index.tsx
```diff
@@ -10,11 +10,13 @@ import Icon from '../Icon';
 import ScrollBox from '../ScrollBox';
 import TransitionAnimation from '../TransitionAnimation';
 import StyledModal, { StyledModalWrapper } from './style';
+import ButtonGroup from '../ButtonGroup';
 
 type PropsType = StyledType & {
     show: boolean;
     title: string;
     size?: 'small' | 'medium' | 'large';
+    buttons?: Array<ReactNode>;
```
CarloPalinckx: We could also limit the amount of buttons that can be used here (if that's something we want to do?) by doing:
```ts
buttons?: [ReactNode, ReactNode] | [ReactNode] // Array containing one or two ReactNodes
```

Dreamdealer: I am familiar with the tuples principle, but since UX hasn't set a limit on the amount of buttons here, we should not make these rules ourselves. Using our common sense will be sufficient until UX comes up with a rule for the amount of buttons.

CarloPalinckx: Allright, array will do fine in that case. We can always add the restriction in a later point in time.

PR #294, opened 2018-12-31
Split buttons and fixed button spinner

CarloPalinckx:
### This PR:
Fixes the hardcoded color prop on the spinner inside a button that was introduced in #292. While resolving this bug a few questions came up, including:
- Code splitting ability of button as static properties.
- Memory usage of having multiple component constructors inside a class
- Usage of the Flat button (turns out this should only be used for icon buttons)
- Usage of the CTA button

This pr proposes to solve this by: 

**Breaking changes** 🔥
- `Button.Cta` `Button.Flat` with each `plain` variant are nog reduced to: `Button` and `IconButton`
- `iconAlign` is no longer supported on the `Button`
- Theme was changed accordingly

- `Spinner` in button now correctly uses the correct color again.

**Checklist** 🛡


src/components/Button/index.tsx
```diff
@@ -1,32 +1,128 @@
...
+                    <Box inline padding={[0, 6, 0, 0]}>
+                        <Icon size="medium" icon={props.icon} color={color} />
+                    </Box>
+                )}
+                <span style={{ color }}>{Children.count(props.children) > 0 ? props.children : props.title}</span>
+            </>
+        );
+    },
+})`
+    ${({ theme, variant, compact, disabled, loading }): string => {
```
hborghols: Why does the styling happen here now instead of in `style.tsx`? It's a bit weird to only have a type definition in `style.tsx`.

hborghols:
`src/components/IconButton/style.tsx` is an empty file

src/components/Button/index.tsx
```diff
@@ -1,32 +1,128 @@
...
+
+            &:focus {
+                ${focus}
+            }
+
+            &:active {
+                ${!loading ? active : ''}
+            }
+
+            &::before {
```
hborghols: This pretty much seems to copy the BareButton style, that seems superfluous

src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -378,6 +346,15 @@ const theme: ThemeType = {
             color: grey.base,
         },
     },
+    IconButton: {
```
hborghols: So much better than having the humongous type for the Button :)

src/components/Button/index.tsx
CarloPalinckx: I used the `styled.attrs` constructor to avoid an extra wrapping. Would you rather prefer the theme definition in the `index.tsx`?

src/themes/MosTheme/MosTheme.theme.ts
CarloPalinckx: 🌈

src/components/Button/index.tsx
hborghols: I'm not sure. It would make `index.tsx` even bigger which is also not ideal. I suppose it's fine like this for now. Because other solutions such as introducing a separate file (`theme.ts` or something) is also a bit weird. If you want to use this way of formatting components more in the future it might be a good idea to open an RFC so we can make decision on how to keep a consistent codebase.

CarloPalinckx: I don't think big files are necessarily a bad thing (https://twitter.com/dan_abramov/status/1008131488481730561 😁). It doesn't really hurt navigation either imo because it is easily foldable in most editors :)

hborghols: Sure, I also don't think big files are that bad (up to a certain point obviously). I just think we should be consistent in how we do things :).

PR #298, opened 2019-01-02
Added a horizontal rendering for EmptyState

CarloPalinckx:
### This PR:

Makes the EmptyState a bit more versatile in small/short containers.

- Added a `horizontal` prop to `EmptyState` which renders it horizontally.
- Added an `illustration` prop which allows you to choose an illustration
- `balloon` illustration added.
-  added `ReactNode` as an eligable type for title, message and illustration that allows you to inverse control on the rendering of those.
- added a `button` prop that does this as well for the button.

- Illustrations now shrink when their container shrinks.

**Checklist** 🛡


CarloPalinckx:
@LuukHorsmans

src/components/EmptyState/test.tsx
```diff
@@ -0,0 +1,15 @@
...
+    it('should render its children', () => {
+        const component = mountWithTheme(
+            <EmptyState title="Foo" message="Bar">
+                <div data-testid="foo" />
+            </EmptyState>,
+        );
+
+        expect(component.find('[data-testid="foo"]').length).toBe(1);
+    });
+});
```
nicokoelewijn: This test doesn't prove anything apart from rendering. 
You could check if the horizontal prop is processed correctly: 

```
    it('should render in a vertical fashion', () => {
        const component = mountWithTheme(<EmptyState title="foo" message="bar" />);

        expect(
            component
                .find(Box)
                .first()
                .prop('direction'),
        ).toEqual('column');
    });

    it('should render in a horizontal fashion', () => {
        const component = mountWithTheme(<EmptyState horizontal title="foo" message="bar" />);

        expect(
            component
                .find(Box)
                .first()
                .prop('direction'),
        ).toEqual('row');
    });
```

src/components/EmptyState/index.tsx
```diff
@@ -1,37 +1,58 @@
...
+            </Box>
+            <Box margin={[12, 0, 24, 0]}>
+                <Text textAlign="center" descriptive>
+                    {props.message}
+                </Text>
+            </Box>
+            {props.children}
+        </Box>
+    );
+};
```
nicokoelewijn: This structure causes a lot of redundant code. This can easily be refactored.

src/components/EmptyState/test.tsx
CarloPalinckx: > This test doesn't prove anything apart from rendering.
Do you mean that you object to the first test itself or that the test suite is incomplete as a whole?

nicokoelewijn: I noticed that the test suite is incomplete, but that's for another PR. My issue is with the first test that doesn't test any functionality.

src/components/EmptyState/index.tsx
CarloPalinckx: Could you elaborate a bit more on how you would like to see this refactored? When looking at both these cases it looks to me that almost every component has different prop values or different nesting.

src/components/EmptyState/test.tsx
CarloPalinckx: I have pushed a commit that contains all of this thread: https://github.com/MyOnlineStore/bricks/pull/298/commits/7627ede832a6c347c4608c73f9bd5b9f6d60b76a

LuukHorsmans:
The standard title size should be h2 for an EmptyState to better fit in the page hierarchy.

PR #321, opened 2019-01-21
Table responsiveness

Dreamdealer:
### This PR:

Added responsive view (datacards) on the table component.

**Breaking changes** 🔥
- None.

- Component `Table` now has a new attr `view` which takes either `table` or `datacard` to switch between views.

- Sorting is "remembered" between switching views
- Width setting of columns is now supported
- Width can be set to `fit` to have the columns as small as possible (used for buttons column)
- Buttons are now set as separate items, not as a regular column
- Buttons are always rendered in the last column
- Table now has a `preSort` prop to set the sorting before loading (and is used to switch between views while keeping the current column sorted

**Checklist** 🛡


CarloPalinckx:
tagging @LuukHorsmans for a design-review :)

src/components/Table/index.tsx
```diff
@@ -1,28 +1,42 @@
...
-    header?: ReactNode;
+    header?: ReactNode | string | number | any;
     align?: 'start' | 'center' | 'end';
+    width?: string;
     sort?(cellA: GenericCellType, cellB: GenericCellType): number;
     render?(cell: GenericCellType, row: GenericRowType): JSX.Element;
 };
 
 type PropsType<GenericRowType extends BaseRowType> = {
+    view?: 'table' | 'datacard';
```
CarloPalinckx: Styled components added an `as` prop for this kind of behavior in v4 (which was updated in #227) maybe it could be nice to as that convention here as well, since we already do it for Text/Heading etc. Also, maybe "card" is a better wording here since data is a but implied when rendering a table :)

```suggestion
   as?: 'table' | 'card';
```

src/components/Table/index.tsx
```diff
@@ -1,28 +1,42 @@
...
+import { MediumIcons } from './../Icon/types';
+import SeverityType from '../../types/_SeverityType';
 
 type SortDirectionType = 'ascending' | 'descending' | 'none';
 
 type BaseRowType = {
     id: string;
     selected?: boolean;
+    buttons?: Array<ReactNode>;
+    statusIcons?: Array<StatusIconType>;
```
CarloPalinckx: I think this can be removed since it is unused.

src/components/Table/story.tsx
```diff
@@ -32,16 +51,64 @@ class Demo extends Component<PropsType, StateType> {
...
-                { id: '61651323', price: 0.8, name: 'Kiwi', image: '🥝', actions: true },
-                { id: '61651320', price: 3.5, name: 'Pineapple', image: '🍍', actions: true },
-                { id: '61651322', price: 2.3, name: 'Grapes', image: '🍇', actions: true },
-                { id: '61651321', price: 1.2, name: 'Banana', image: '🍌', actions: true },
-                { id: '61651324', price: 0.7, name: 'Lemon', image: '🍋', actions: true },
+                {
+                    id: '61651323',
+                    price: 0.8,
+                    name: 'Kiwi',
+                    image: <img src="https://via.placeholder.com/50" width="50" />,
```
CarloPalinckx: Would be cool if you could add actual images of fruit :)

src/components/Table/story.tsx
```diff
@@ -1,16 +1,34 @@
 import { storiesOf } from '@storybook/react';
-import React, { Component, Fragment } from 'react';
+import React, { Component, Fragment, ReactNode } from 'react';
```
CarloPalinckx: Overall this story has a lot of new usage added. Maybe you could dress it down a bit? I think showing how to render a different color for the price is plenty for the custom rendering for now. Custom rendering is a bit of a escape hatch for now, I wouldn't encourage people too much to go wild with custom rendering 😄

src/components/Table/story.tsx
```diff
@@ -54,80 +121,140 @@ class Demo extends Component<PropsType, StateType> {
...
+        );
+    };
+
+    private renderImage = (image: string | ReactNode) => {
+        return <>{image}</>;
+    };
+
+    private renderActions = (id: number) => {
+        return [
+            // tslint:disable:jsx-wrap-multiline
```
CarloPalinckx: Does this tslint rule conflict with out prettier settings?

src/components/Table/story.tsx
```diff
@@ -1,16 +1,34 @@
 import { storiesOf } from '@storybook/react';
-import React, { Component, Fragment } from 'react';
+import React, { Component, Fragment, ReactNode } from 'react';
 import Table from '.';
 import Text from '../Text';
-import { boolean } from '@storybook/addon-knobs';
 import IconButton from '../IconButton';
+import { boolean, select } from '@storybook/addon-knobs';
```
CarloPalinckx: Few of these imports are unused.

src/components/Table/test.tsx
```diff
@@ -5,19 +5,99 @@ import { DragDropContext } from 'react-beautiful-dnd';
...
+            />,
+        );
+
+        expect(table.find(Card).length).toBe(4);
+    });
+
+    it('should render a column with buttons when buttons are provided', () => {
+        const table = mountWithTheme(
             <Table
+                view={'table'}
```
CarloPalinckx: You can just use view="table" here

src/components/Table/Card/test.tsx
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+import Card from '.';
+import { mountWithTheme } from '../../../utility/_styled/testing';
+import StyledRow from './style';
+import { ContrastThemeProvider } from '../../Contrast';
+
+describe('Cards', () => {
+    it('should handle mouse focus and blur when draggable', () => {
```
CarloPalinckx: This test has no expect :)

src/components/Table/test.tsx
```diff
@@ -5,19 +5,99 @@ import { DragDropContext } from 'react-beautiful-dnd';
...
+            table
+                .find(Row)
+                .first()
+                .find(Cell)
+                .last()
+                .find(Button).length,
+        ).toBe(2);
+    });
+
+    it('should render a CompactHeader with a Select when there are column with sort', () => {
```
CarloPalinckx: Maybe you could expand this test a bit, would be nice to test if the sorting can be applied with the select as well.

src/components/Table/Card/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+import StyledCard from './style';
+import Checkbox from '../../Checkbox';
+import Icon from '../../Icon';
+import Branch from '../../Branch';
+import { Draggable, DraggableProvided } from 'react-beautiful-dnd';
+import { ContrastThemeProvider } from '../../Contrast';
+
+type PropsType = {
+    // tslint:disable
+    columns: {
```
CarloPalinckx: you can use `// tslint:disable-next-line` here instead of the double comments

src/components/Table/Card/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+                                    return (
+                                        (this.props.columns[a].order as number) -
+                                        (this.props.columns[b].order as number)
+                                    );
+                                })
+                                .map((key, index) => {
+                                    const column = this.props.columns[key];
+                                    const cell = this.props.row[key];
+                                    const align = column.align ? column.align : 'start';
```
CarloPalinckx: align is unsed here.

src/components/Table/CompactHeaders/index.tsx
```diff
@@ -0,0 +1,179 @@
+import React, { Component } from 'react';
+import Box from '../../Box';
+import Checkbox from '../../Checkbox';
+import { ColumnType, SortDirectionType } from '..';
+import Select from '../../Select';
+
+enum SortingSteps {
```
CarloPalinckx: This enum is unused.

src/components/Table/Card/style.tsx
```diff
@@ -0,0 +1,57 @@
...
+    focus?: boolean;
+};
+
+const StyledCard = withProps<StyledRowProps>(styled.div)`
+    transition: background-color 300ms, border 300ms;
+    border-radius: 3px;
+    box-shadow: 0 2px 10px 0 rgba(33, 37, 43, .15);
+    margin-bottom: 24px;
+
+    table {
```
CarloPalinckx: We try to avoid contextual styling as much as possible, would you be so kind to refactor these to styled components?

src/components/Table/CompactHeaders/index.tsx
```diff
@@ -0,0 +1,179 @@
...
+            };
+        });
+
+        this.setState({
+            columns,
+        });
+    };
+
+    public handleChange = (key: string): void => {
+        let direction: SortDirectionType = 'descending';
```
CarloPalinckx: We try to avoid mutability on variables. You can turn these into `const` by doing a ternary on the assignment:
```ts
const direction = key.indexOf('_ascending') > 0 ? 'ascending' : 'descending' ;
``` 
etc.

LuukHorsmans:
Great work! I really like the way this works for our tables. Just a few tweaks to help with the visual hierarchy and upcoming changes to other components. 

- Please remove the green overlays. The green does not work well with the greys used in the tables and there were some technical challenges with the grouped border. So for now: skip the green overlays.
- In the responsive view/cardview please change the labels/titles to a bold font-weight to create visual hierarchy.
- Use icons in the sorting dropdown instead of A-Z / Z-A for Pricing
- I agree with Carlo regarding the custom rendering. I'd stick to the custom pricing and maybe the sticker badges. I would remove the icons and the custom array to discourage too many 'custom' implementations (or atleast change the way it is presented in storybook). Other than that: please change the label in the cardview to align with the first line of the product (product <--> Banana).

src/components/Table/index.tsx
Dreamdealer: Okay, good suggestion. I changed the `view` prop to `as` everywhere.

src/components/Table/story.tsx
LuukHorsmans: I agree. Maybe we can change the way the custom rendering is presented in storybook?

src/components/Table/index.tsx
Dreamdealer: True, i overlooked that one.

src/components/Table/story.tsx
Dreamdealer: Haha, I've looked for a placeholder image service that serves fruit, but I went with these generic placeholders instead.

src/components/Table/story.tsx
Dreamdealer: Yeah, i've talked about this with Luuk this morning and I am going to remove all teh custom rendering except for the badges and prices.

src/components/Table/story.tsx
Dreamdealer: I removed these lines and there aren't any errors anymore. Looks like a leftover from previous versions before the final rebase. (when buttons and icons weren't combined in the IconButton component)

src/components/Table/story.tsx
Dreamdealer: Ah, also leftovers after the last rebase. I removed the unused imports.

src/components/Table/CompactHeaders/index.tsx
Dreamdealer: Yeah, I've been struggling with this one. Thanks for the example.. I now get it and refactored to get rid of the let's.

src/components/Table/Card/style.tsx
Dreamdealer: I think I need a little bit of learning on this subject... I'll try to refactor this!

src/components/Table/test.tsx
Dreamdealer: Omg this was a tricky one (triggering the filter), but I managed to write a test for it.

src/components/Table/index.tsx
```diff
@@ -1,28 +1,35 @@
...
-import Headers from './Headers';
+import { MediumIcons } from './../Icon/types';
+import SeverityType from '../../types/_SeverityType';
 
 type SortDirectionType = 'ascending' | 'descending' | 'none';
 
 type BaseRowType = {
     id: string;
     selected?: boolean;
+    buttons?: Array<ReactNode>;
```
CarloPalinckx: I would advise against putting ReactNodes in the rows because it reduces the change that you can directly use data from an external data source. In a real world scenario you would end up having to map the dataset, which doesnt have to be a bad thing. But we already expose a way to do this with the render property on the ColumnType. ReactNode are also pretty awkward to update once put in the state. If you want to update an image based on width for instance, you would have to put a piece of JSX in a setState call. Which feels a bit icky compared to using custom rendering for this.

CarloPalinckx: For the buttons this could also make the implementation easier because you dont have a pass it as a prop to the Row but just sniff out if there is render defined for the 'buttons' column directly in the Row.

src/components/Table/index.tsx
```diff
@@ -1,28 +1,35 @@
...
     // tslint:disable-next-line
-    [key: string]: string | number | boolean | undefined;
+    [key: string]: string | number | boolean | undefined | Array<ReactNode> | ReactNode;
```
CarloPalinckx: Same applies here.

src/components/Table/story.tsx
```diff
@@ -23,6 +39,7 @@ type PropsType = {
     selectable: boolean;
     sortable: boolean;
     custom: boolean;
+    as?: 'datacard' | 'table';
```
CarloPalinckx: datacard can be card here :)

src/components/Table/index.tsx
```diff
@@ -1,28 +1,35 @@
...
-    header?: ReactNode;
+    header?: ReactNode | string | number | any;
     align?: 'start' | 'center' | 'end';
+    width?: string;
     sort?(cellA: GenericCellType, cellB: GenericCellType): number;
     render?(cell: GenericCellType, row: GenericRowType): JSX.Element;
 };
 
 type PropsType<GenericRowType extends BaseRowType> = {
+    as?: 'table' | 'datacard';
```
CarloPalinckx: I guess "data" is implied for a Table, I would turn `datacard` into `card`

src/components/Table/Card/index.tsx
```diff
@@ -0,0 +1,174 @@
...
+                        </Box>
+                    )}
+                    {this.props.draggable && provided && (
+                        <Box padding={[6]} {...provided.dragHandleProps}>
+                            <Text severity={!this.state.hasHover ? 'info' : undefined}>
+                                <Icon size="medium" icon="bars" />
+                            </Text>
+                        </Box>
+                    )}
+                    {Object.keys(this.props.columns).map(key => {
```
CarloPalinckx: You can skip the `Object.keys` here because you know the index of the column :) 

You can just do `if (this.props.columns.buttons !== undefined) ...` here

src/components/Table/Row/index.tsx
```diff
@@ -11,8 +11,14 @@ import Branch from '../../Branch';
 import { ColumnType, BaseRowType } from '..';
 
 type PropsType = {
-    // tslint:disable-next-line
-    columns: { [key: string]: ColumnType<string | number | boolean | undefined, any> };
+    // tslint:disable
+    columns: {
+        [key: string]: ColumnType<
+            string | number | boolean | undefined | Array<ReactNode> | ReactNode | Array<StatusIconType>,
```
CarloPalinckx: I think this change can be reverted as well now that you check for buttons inside the Row.

src/components/Table/story.tsx
```diff
@@ -1,16 +1,32 @@
...
 type RowType = {
+    selected?: boolean;
     id: string;
     price: number;
     name: string;
     image: string;
-    actions: boolean;
+    badge?: string;
+    extra?: string;
+    buttons?: Array<ReactNode>;
```
CarloPalinckx: A few extraneous props left here

src/components/Table/Card/index.tsx
Dreamdealer: I have no idea what I was thinking there... :retard:

Dreamdealer:
Tested in Browserstack on:
1. IE Edge - windows 10 > OK
2. IE 11 - windows 8.1 > JS fail
3. Opera 58 - windows 10 > OK
4. Yandex 14.12 - High Sierra > Storybook fail
5. Firefox 65 - High Sierra > OK
6. Safari 11.1 - High Sierra >OK

Issue #313, opened 2019-01-14
Set focus styles on flat Buttons


woutsluiter:
### Bricks version 📦
`0.9.2`

### Describe the bug 🔍
When global css properties are inherited on the focus state of a `<Button variant="plain" />`. Seems that `FlatButtonStylesType` doesn't have a focus state.

### To Reproduce 💥
- Create global focus styling on a `button` element
- Add a `<Button variant="plain" />`
- Focus

### Expected behavior ✨
The styles on a focus state are set as well.

### Screenshots 📸
![image](https://user-images.githubusercontent.com/10552400/51107090-1b594000-17ee-11e9-92aa-3dab3c5a235e.png)

### Context👨‍💻
 - Device: desktop
 - OS: OSX
 - Browser Chrome
 - Browser version 71.0.3578.98

CarloPalinckx:
Since this can't be fixed without a breaking fix, and the problem should no longer occur in the upcoming version I'm closing this.

woutsluiter:
It actually still a thing in the 1.x branch. The IconButton has no focus styling and inherits common globally styled properties like `background` on hover and active states. I would be happy to make a PR for this. Could be something like the way it's done in `Button`:

```
type CommonType = {
    backgroundColor: string;
    color: string;
};

type ComponentStateTypes = {
    idle: CommonType;
    hover: CommonType;
    focus: CommonType;
    active: CommonType;
};

type IconButtonThemeType = {
    primary: ComponentStateTypes
    destructive: ComponentStateTypes
};
```

woutsluiter:
Resolved in #330

PR #330, opened 2019-02-07
Explicit state styling on (Icon)Button

woutsluiter:
### This PR:

Resolves: #313

The `IconButton` in the `1.x` branch had no focus state styling. Therefor it inherited styling which is commonly defined in applications, like `background` and `color`. It now has explicit styling for hover, focus, and active states. This inludes a fallback to `idle` when in a `loading` state, which was undefined before. This is fallback is also applied to `Button`.

**Breaking changes** 🔥
- Adjusted IconButtonThemeType

- Button & IconButton have explicit styling

**Checklist** 🛡

### DIFF
--- a/src/themes/MosTheme/MosTheme.theme.ts
+++ b/src/themes/MosTheme/MosTheme.theme.ts
@@ -339,37 +339,17 @@ const theme: ThemeType = {
     IconButton: {
         primary: {
             idle: {
-                backgroundColor: 'transparent',
-                color: grey.base,
+                color: grey.lighter1,
             },
             hover: {
-                backgroundColor: 'transparent',
-                color: grey.darker1,
-            },
-            focus: {
-                backgroundColor: 'transparent',
-                color: grey.darker1,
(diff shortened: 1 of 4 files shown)


src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -338,11 +338,40 @@ const theme: ThemeType = {
     ],
     IconButton: {
         primary: {
-            color: grey.darker1,
-            backgroundColor: 'transparent',
+            idle: {
```
hborghols: seems a bit excessive to have the same styles copied everywhere. I would just use the same variable but set them explicitly in the style.tsx

hborghols:
hover/active/etc styles are still applied in disabled state

src/themes/MosTheme/MosTheme.theme.ts
hborghols: Much better 😄

PR #338, opened 2019-02-20
Improved keyboard navigation

hborghols:
### This PR:

To (slighty) improve a11y I checked which components are usable with only the keyboard. All of them were apart from `Table` and there was a slight bug in `Range` when using the keyboard. Both of these issues are fixed in this PR.

- It is now possible to sort a `Table` using only the keyboard
- It is no longer possible to enter a value outside of the min/max limit on the Range using the keyboard

**Checklist** 🛡


CarloPalinckx:
👏

PR #339, opened 2019-02-21
Checkbox label

CarloPalinckx:
### This PR:

Adds the ability to render a label next to a Checkbox

- Component `Checkbox` now also supports an optional `label` prop.

**Checklist** 🛡

### DIFF
--- a/src/components/Checkbox/index.tsx
+++ b/src/components/Checkbox/index.tsx
@@ -16,8 +16,7 @@ type PropsType = {
     name: string;
     label?: string;
     id?: string;
-    // tslint:disable-next-line:no-any
-    onChange(change: { checked: boolean | 'indeterminate'; event: MouseEvent<any> }): void;
+    onChange(change: { checked: boolean | 'indeterminate'; event: MouseEvent<HTMLDivElement> }): void;
 };
 
 class Checkbox extends Component<PropsType, StateType> {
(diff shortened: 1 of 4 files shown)


src/components/Checkbox/story.tsx
```diff
@@ -1,4 +1,4 @@
-import { boolean } from '@storybook/addon-knobs';
+import { boolean, text } from '@storybook/addon-knobs';
```
hborghols: Would be dope to also finally make the checkbox clickable in the story

src/components/FormRow/story.tsx
```diff
@@ -37,6 +40,8 @@ class DemoComponent extends Component<PropsType, StateType> {
             city: '',
             country: '',
             toggled: false,
+            bacon: false,
+            cheese: true,
```
hborghols: 👌

src/components/Checkbox/index.tsx
```diff
@@ -13,8 +14,10 @@ type PropsType = {
     error?: boolean;
     value: string;
     name: string;
+    label?: string;
     id?: string;
-    onChange(change: { checked: boolean | 'indeterminate'; event: MouseEvent<HTMLDivElement> }): void;
+    // tslint:disable-next-line:no-any
+    onChange(change: { checked: boolean | 'indeterminate'; event: MouseEvent<any> }): void;
```
hborghols: Why is it necessary to introduce an `any` here?

CarloPalinckx: It isn't, good call.

PR #341, opened 2019-02-22
Hover fix responsive tables

Dreamdealer:
### This PR:

Fixes incorrect hover state on Table card view

**Bugfixes and internal changes** 🎈
- Optional prop `hover` added. When prop is not defined or true the table functions the same as before.

**Checklist** 🛡


src/components/Table/Row/style.tsx
```diff
@@ -19,7 +20,8 @@ const StyledRow = styled.tr<PropsType>`
     outline: ${({ focus, theme }): string => (focus ? `solid 4px ${theme.Table.row.focus.borderColor}` : '')};
 
     &:hover {
-        background-color: ${({ theme }): string => theme.Table.row.hover.backgroundColor};
+        background-color: ${({ hover, theme }): string =>
+            hover !== false ? theme.Table.row.hover.backgroundColor : '#FFFFFF'};
```
CarloPalinckx: shouldn't #fff be transparent here? The background isn't guaranteed to be white

Dreamdealer: No, the overall background color of the table is grey and on hover the row becomes transparent this is due to the contrast component on the table. To counteract this I needed to explicitly set the hover color to white, which is the non-hover color for a row also.

woutsluiter: Can't you fallback on the default theme backgroundColor then? Nicer than a hardcoded color IMO.

Dreamdealer: You're completely right! Didn't think about that option. Will make that asap.

Dreamdealer: It has been done.

Dreamdealer:
The `ContrastThemeProvider` used in the card was indeed not used and therefor I removed it, but this didn't resolve the issue and the need for my previous fix.

The regular Table component is also used inside the new cards. So the unwanted grey background hover in the cards is not caused by the card but by the nested table inside.

PR #350, opened 2019-03-12
Added debounce to Range

CarloPalinckx:
### This PR:

- Debounced onChange calls for Range

**Checklist** 🛡

### DIFF
--- a/src/components/Range/index.tsx
+++ b/src/components/Range/index.tsx
@@ -98,7 +98,7 @@ class Range extends Component<PropsType, StateType> {
     };
 
     // tslint:disable-next-line
-    private handleChange = debounce(16, (values: RangeType): void => {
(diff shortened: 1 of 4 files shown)


src/components/Range/index.tsx
```diff
@@ -96,6 +97,13 @@ class Range extends Component<PropsType, StateType> {
         if (this.props.onChange !== undefined) this.props.onChange(inputValues());
     };
 
+    // tslint:disable-next-line
+    private handleChange = debounce(16, true, (values: RangeType): void => {
```
nicokoelewijn: 16ms makes the dragging of the sliders feel janky. A value of 6 would be a better fit. 
```suggestion
    private handleChange = debounce(6, true, (values: RangeType): void => {
```

PR #355, opened 2019-03-25
Component theme fixes for 1.x

Dreamdealer:
### This PR:

Fixes a few styling issues that occurred when trying to them all the bricks components.

**Breaking changes** 🔥
- None

- New theme vars are added to default MosTheme for backwards compatibility.

- Added vertical labels option to Formrow component (forces labels and form elements on individual lines)
- Added themable checkmark color to Radiobutton and Checkbox
- Link component and textarea component now inherits font-family from parent element
- Textarea font color is now themable

**Checklist** 🛡


src/components/FormRow/index.tsx
```diff
@@ -1,20 +1,34 @@
 import React, { ReactNode, FunctionComponent } from 'react';
 import Box from '../Box';
-import trbl from '../../utility/_trbl';
 import { StyledFormRow } from './style';
 
 type PropsType = {
     label: ReactNode;
     field: ReactNode;
+    vertical?: boolean;
```
hborghols: What is this for?

Dreamdealer: As mentioned in the PR description: `Added vertical labels option to Formrow component (forces labels and form elements on individual lines)`

Dreamdealer: As seen in this screenshot:

image

hborghols: Ah sorry, missed that

hborghols: But why does this have to be themeable?

src/components/TextField/style.tsx
```diff
@@ -51,6 +51,8 @@ const StyledInput = styled.input<InputPropsType>`
     margin: 0;
     background: ${({ theme, disabled }): string =>
         disabled ? theme.TextField.disabled.background : theme.TextField.idle.common.background};
+    color: ${({ theme }): string => theme.TextField.idle.common.color};
```
hborghols: This makes the text look completely different from how it does now. Is there an indication the text color should be themeable?

src/components/FormRow/index.tsx
Dreamdealer: It is not actually a theme related addition, but more of a functionality I missed when setting up my testcase.

src/components/TextField/style.tsx
Dreamdealer: Have you got a screen of what you're seeing? In my test setup this works fine..

src/components/FormRow/index.tsx
hborghols: Sorry but I don't really see the need for this functionality. If you think it's important to add it could you open a RFC (feature request)? In any case I think it's out of the scope of this PR.

src/components/TextField/style.tsx
hborghols: With completely different I meant the text color btw 😄

hborghols: This branch:
Screenshot 2019-03-25 at 13 43 13

1.x branch:
Screenshot 2019-03-25 at 13 43 20

Dreamdealer: Hmm, I see. Without having a look at the problem I guess the behaviour in 1.x is actually a bug. But i'll make the text color optional.

src/components/FormRow/index.tsx
Dreamdealer: I will remove it.

src/components/TextField/style.tsx
hborghols: Master also looks the same so I don't know if it's a bug but I don't think it's necessary to make it optional, just making it look the same as it does now is fine :)

Dreamdealer: This is a problem. I've looked into it and apparently the color in master/1.x is #000. Which is not a color used in our MOS theme. The setting which I have applied implies that this should have been `grey.lighter3` (#A6AAB3) which should be #333740 according to the designs in abstract (see screen).

I am going to change the A6AAB3 color setting to 333740. The difference in color is so small that I don't think this will cause any problems.

image

Dreamdealer: image

image

Dreamdealer: I have notified Luuk for this change. I can't really imagine that this change will be a problem, but I'd rather be sure.

hborghols: 👍 👍

PR #356, opened 2019-03-26
Update react(-dom) versions to be able to use Hooks

Dreamdealer:
### This PR:

Updates React to enable Hooks.

**Breaking changes** 🔥
- None.

- None.

- React(-dom) version 16.7 > 16.8

**Checklist** 🛡


src/components/Checkbox/index.tsx
```diff
@@ -19,7 +19,7 @@ type PropsType = {
     label?: string;
     id?: string;
     'data-testid'?: string;
-    onChange(change: { checked: boolean | 'indeterminate'; event: MouseEvent<HTMLDivElement> }): void;
+    onChange(change: { checked: boolean | 'indeterminate'; event: React.MouseEvent<HTMLDivElement> }): void;
```
woutsluiter: I think the problem is that the methods in this file are public, where they should be private.

Dreamdealer: You're right... I have no idea what I was doing here...

Dreamdealer:
@woutsluiter package lock is added

package.json
```diff
@@ -91,10 +91,10 @@
     "webpackbar": "^3.1.4"
   },
   "peerDependencies": {
-    "@types/react": "^16.7.6",
-    "@types/react-dom": "^16.0.9",
-    "react": "^16.6.3",
-    "react-dom": "^16.6.3",
+    "@types/react": "^16.8.5",
```
CarloPalinckx: the deps can all be `^16.8.0` (everything higher than 16.8.0, but not past 17.x.x) they don't need the specific .5 patches to work. If we find bugs with that configuration because of a missing patch we can bump the requirements later on.

Dreamdealer: I wasn't sure which version to use exactly, so i used the latest version i found on npmjs

PR #357, opened 2019-03-27
Added focus border and boxshadow for textarea and textfield

Dreamdealer:
### This PR:

For the focus state of textfields and textareas the severity color scheme's were used. Since the severity colors match our MOS theme colors this was never a problem, untill the merchant theme's could have non-green primary colors.

So I changed the behaviour of the text(area/field)s to use a seperate color settings for the focus border and boxshadows. It is still possible to use the 'success' severity type.

For the MOS theme nothing changes, because the colors are the same, but this enables us to set different colors in the merchant themes.

**Breaking changes** 🔥
- None

- None

- Separated focus colors from severity colors.

**Checklist** 🛡


Dreamdealer:
> Looks good, would you mind having a look at all the usages of the `Text` subtheme here? These should never be used cross components. Not sure how they got in here in the first place 😅

I saw that, but figured it was outside the scope of this fix. But it was an easy fix and I added it anyways.

PR #379, opened 2019-05-15
Added yarn and updated node image 🐈🧶

CarloPalinckx:
### This PR:

Adds yarn and updates the node image, for some reason I could not properly build the master without updating the node types.


CarloPalinckx:
> Is the yarn.lock up to date 

I just updated it to the latest, the equivalent for the --no-save flag in npm is `--frozen-lockfile`
So:
`npm i --no-save` vs `yarn --frozen-lockfile`

PR #392, opened 2019-06-06
Add prop tables / component API documentation

### LINKED ISSUE #390: Add overview of PropTypes for each component to storybook
Currently we don't have any documentation showing the API of our components. To get an overview of which props are available, required and what values they take it's necessary to dive into the source code or rely on typeerrors, both of which work but are not the most user-friendly.

I made a small PoC that uses https://github.com/storybookjs/storybook/tree/next/addons/info and https://github.com/strothj/react-docgen-typescript-loader/ to be able to generate tables of PropTypes for components. You can find this in the [poc-proptables branch](https://github.com/MyOnlineStore/bricks/tree/poc-proptables).

I propose adding a separate `PropsType` or `Proptable` story for each component that uses the info plugin with the `inline` option set to `true` to immediately show a table of props. This also requires adding descriptions with additional information to the type definitions. Adding a story with info settings is as simple as:

```
storiesOf('Modal', module).add(
   'Props table',
   withInfo({ inline: true })(() => {
       return <Modal show={false} size="medium" title="Props table!" />;
   }),
);
```
Adding additional descriptions to the type definition simply uses comments:
```
type PropsType = {
    /** Determines whether the modal is visible or not */
    show: boolean;

    /** Title shown at the top of the modal */
    title: string;

    etc...
```
Descriptions are not necessary for all props but are useful for some, e.g., it's not necessary to explain `show: boolean;` but it may be useful to add some context to `renderFixed?(): JSX.Element`.

- Add an info tab to all existing stories. This is easily possible by adding `addDecorator(withInfo)` to the storybook `config.js` file. However, we have quite a few elaborate demo stories which are not suitable for generating a proptable. The plugin generates a table for each element used in the story so it's clearest is the story *only* uses the element we want to document.
- Add a clickable info tab to current stories where suitable and add new ones where necessary. I personally prefer the inline view and it seems like this would still require adding quite a few new stories to be able to show the tables properly. Converting current stories may be possible with the proposed solution anyway.
- There doesn't really seem to be a good alternative way to generate prop tables in storybook. I haven't looked into options outside of storybook yet.

### Additional challenges
- It may be necessary to filter out some of the props for some of the components. This appears to be possible using the docgen-loader but I haven't looked into this in depth yet. 

### Disadvantages:
- The docgen plugin makes the webpack build significantly slower.

### Additional context 🖼
Example table that I generated in the PoC:
Screenshot 2019-06-03 at 16 25 22

hborghols:
### This PR:

resolves #390 #9 
- Adds Prop tables in the `Default` story of a component story if suitable and in a separate `Props` story if not.
- To facilitate this `@storybook/addon-info` and `react-docgen-typescript-loader` are added.

I'm still running into one issue: the prop tables of `Button (and IconButton and MultiButton), Select, Heading and Text` can't be generated because they directly use the styled components `styled.element` thing which somehow doesn't work. A potential solution could be (re-)wrapping the elements into e.g., `StyledButton` and `Button = (props) => <StyledButton />`, like I've done with `Badge`. However before I try to make that work I'd like some input on potential alternative solutions.

Aditionally I'd like some feedback on the `Box` props table. We define the props as: `type PropsType = JSX.IntrinsicElements['div'] & {...restOfProps};` so theoretically all `div`'s intrinsic props are part of our type definition. This leads to a gigantic props table however.

Finally: the additional of `react-docgen-typescript-loader` slows down the webpack process quite a bit. Please check if it's still acceptable. Otherwise my proposal would be to only add prop tables in static builds by default and to add a separate flag/script to run the dev environment with prop tables.

TODO:
- [x] Add descriptions to PropTypes where necessary
- [ ] Address issues outlined above

**Checklist** 🛡


hborghols:
Closed for now due to the developments regarding a static docs site

PR #400, opened 2019-07-02
Select fix scroll

hborghols:
### This PR:

Fixes the behavior of the `Select` window. It now updates on scroll 🌈 .

- `Select` window position is now updated on scroll events. For now these are throttled by `window.requestAnimationFrame()` as suggested in https://developer.mozilla.org/en-US/docs/Web/API/Document/scroll_event . I actually tried using `throttle-debounce` but the throttle delay needed to be so short for it to look smooth that the performance gains over requestAnimationFrame seemed negligible.

**Checklist** 🛡
- [ ] I have tested my addition in all supported browsers and for responsiveness (Chrome, Firefox, Safari, Edge, IE11 and mobile browsers).

### DIFF
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -163,7 +163,6 @@ class Select<GenericOptionType extends OptionBaseType> extends Component<PropsTy
     };
 
     private updateWindowTop = (): void => {
-        console.debug('updating top');
         if (
             this.windowRef &&
             this.windowRef.current &&
@@ -182,7 +181,9 @@ class Select<GenericOptionType extends OptionBaseType> extends Component<PropsTy
     };
 
     private handleScroll = (): void => {
-        window.requestAnimationFrame(this.updateWindowTop);
+        if (this.state.isOpen) {
+            window.requestAnimationFrame(this.updateWindowTop);
+        }
     };
 
     public componentDidUpdate(_: PropsType<GenericOptionType>, prevState: StateType): void {
--- a/src/components/Select/index.tsx
+++ b/src/components/Select/index.tsx
@@ -45,6 +45,7 @@ type PropsType<GenericOptionType extends OptionBaseType> = {
     options: Array<GenericOptionType>;
     emptyText: string;
     disabled?: boolean;
+    topOffset?: number;
     'data-testid'?: string;
     onChange(value: string): void;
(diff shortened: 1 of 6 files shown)


hborghols:
Mobile safari 💩

hborghols:
> Mobile safari 💩

Turns out to be a storybook issue rather than an issue with this PR

src/components/Select/index.tsx
```diff
@@ -153,6 +161,24 @@ class Select<GenericOptionType extends OptionBaseType> extends Component<PropsTy
...
+            this.state.inputHeight
+        ) {
+            const newTop = this.wrapperRef.current.offsetTop + this.state.inputHeight + windowInnerOffset;
+
+            this.windowRef.current.style.top = `${newTop.toString()}px`;
+        }
+    };
+
+    private handleScroll = (): void => {
+        window.requestAnimationFrame(() => this.updateWindowTop());
```
woutsluiter: Don't know if it has a real impact on performance, but you could just pass the function instead of wrapped in an anonymous one.

src/components/Select/index.tsx
```diff
@@ -153,6 +161,24 @@ class Select<GenericOptionType extends OptionBaseType> extends Component<PropsTy
...
 
+    private updateWindowTop = (): void => {
+        if (
+            this.windowRef &&
+            this.windowRef.current &&
+            this.wrapperRef &&
+            this.wrapperRef.current &&
+            this.state.inputHeight
+        ) {
+            const newTop = this.wrapperRef.current.offsetTop + this.state.inputHeight + windowInnerOffset;
```
woutsluiter: This calculation will position the window incorrectly when the parent element is positioned lower than the page top, like a div halfway on the page with an overflow auto.

src/components/Select/index.tsx
```diff
@@ -232,6 +270,7 @@ class Select<GenericOptionType extends OptionBaseType> extends Component<PropsTy
                                             ? `${this.props['data-testid']}-input-field`
                                             : undefined
                                     }
+                                    style={{ width: '100%' }}
```
hborghols: This was added as requested by nico. It fixes a bug where there is a discrepancy between the input width and the window width.

PR #401, opened 2019-07-08
Slider component

nicokoelewijn:
### This PR:

Creates a `Slider` component.
This is a component similar to the `Range` component but it only has one value.
This component is required for the https://myonlinestore.atlassian.net/browse/STRF-146 ticket

- The new component `Slider` has been added.
- Component `Box` has a new `alignItems` value: `baseline`

- Fixed minor typo in range css

**Checklist** 🛡

### DIFF
--- a/src/components/Slider/index.tsx
+++ b/src/components/Slider/index.tsx
@@ -50,6 +50,7 @@ const Slider: FC<PropsType> = props => {
             <Box width="100px" shrink={0} padding={[0, 0, 0, 18]}>
                 <TextField.Number
                     value={inputValue}
+                    disabled={props.disabled}
                     name="slider-value"
                     onChange={(value: number) => {
                         if (props.disabled !== true) {
--- a/src/components/Slider/index.tsx
+++ b/src/components/Slider/index.tsx
@@ -3,6 +3,7 @@ import InputRange from 'react-input-range';
 import StyledWrapper from './style';
 import TextField from '../TextField';
 import { Box } from '../..';
+import memoize from 'memoize-one';
 
 type PropsType = {
     minLimit: number;
@@ -15,16 +16,18 @@ type PropsType = {
 
 const Slider: FC<PropsType> = props => {
     const [inputFocus, setInputFocus] = useState(false);
-    const [inputValue, setInputValue] = useState(props.value ? props.value : props.minLimit);
+    const roundValue = memoize(Math.round.bind(Math));
+    const [inputValue, setInputValue] = useState(props.value ? props.value : roundValue(props.value));
 
-    const isWithinRange = (min: number, max: number, value: number): boolean => value <= max && value >= min;
-
-    const onBlurValue = () => {
-        if (inputValue > props.maxLimit) {
-            setInputValue(props.maxLimit);
-        } else if (inputValue < props.minLimit) {
-            setInputValue(props.minLimit);
+    const setValidValue = (value: number) => {
+        if (value > props.maxLimit) {
+            setInputValue(roundValue(props.maxLimit));
+        } else if (value < props.minLimit) {
+            setInputValue(roundValue(props.minLimit));
+        } else {
+            setInputValue(roundValue(value));
         }
+        if (props.onChange !== undefined) props.onChange(value);
     };
 
     return (
@@ -34,9 +37,10 @@ const Slider: FC<PropsType> = props => {
                     value={inputValue}
                     disabled={props.disabled}
                     onChange={(value: number) => {
-                        setInputFocus(false);
-                        setInputValue(value);
-                        if (props.onChange !== undefined) props.onChange(value);
+                        if (props.disabled !== true) {
+                            setValidValue(value);
+                            setInputFocus(false);
+                        }
                     }}
                     minValue={Math.floor(props.minLimit)}
                     maxValue={Math.ceil(props.maxLimit)}
@@ -47,13 +51,11 @@ const Slider: FC<PropsType> = props => {
                 <TextField.Number
                     value={inputValue}
                     name="slider-value"
-                    onBlur={onBlurValue}
-                    onChange={setInputValue}
-                    feedback={
-                        !isWithinRange(props.minLimit, props.maxLimit, inputValue)
-                            ? { severity: 'error', message: '' }
-                            : undefined
-                    }
+                    onChange={(value: number) => {
+                        if (props.disabled !== true) {
+                            setValidValue(value);
+                        }
+                    }}
                 />
             </Box>
         </Box>
--- a/src/components/Slider/index.tsx
+++ b/src/components/Slider/index.tsx
@@ -1,0 +1,63 @@
+import React, { FC, useState } from 'react';
+import InputRange from 'react-input-range';
+import StyledWrapper from './style';
+import TextField from '../TextField';
+import { Box } from '../..';
+
+type PropsType = {
+    minLimit: number;
+    maxLimit: number;
+    disabled?: boolean;
+    value?: number;
+    label?: string;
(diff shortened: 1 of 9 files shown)


hborghols:
Using the arrows in the number input it's possible to go beyond the min/max values

hborghols:
> Using the arrows in the number input it's possible to go beyond the min/max values

This also seems to be true for `Range`

src/components/Slider/style.tsx
```diff
@@ -0,0 +1,126 @@
...
+/* tslint:disable */
+const styles = require('react-input-range/lib/css/index.css').toString();
+/* tslint:enable */
+
+// prettier-ignore
+const sliderStyles = css`
+    ${styles}
+`;
+
+type SliderThemeType = {
```
hborghols: Maybe you can check with UX if they expect this will ever differ from Range, otherwise having the same type twice is superfluous.

src/components/Slider/style.tsx
```diff
@@ -0,0 +1,126 @@
...
+    box-sizing: border-box;
+
+    & {
+        .input-range__track,
+        .input-range__slider-container {
+            ${({ focus }:PropsType): string => (!focus ? 'transition: none;' : '')}
+        }
+
+        .input-range__track {
+            background: ${({ theme }): string => theme.Range.default.track.background};
```
hborghols: Because you use the Range theme here anyway ;).

src/components/Slider/style.tsx
```diff
@@ -0,0 +1,126 @@
...
+            ${({ theme, disabled }): string => (!disabled ? `box-shadow: ${theme.Range.default.active.boxShadow}` : '')}
+        }
+
+        .input-range__label-container {
+            display: none;
+        }
+    }
+`;
+
+const composeSliderTheme = (themeTools: ThemeTools): SliderThemeType => {
```
hborghols: Same here

src/components/Slider/story.tsx
```diff
@@ -0,0 +1,13 @@
+import Slider from '.';
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { number } from '@storybook/addon-knobs';
+
+storiesOf('Slider', module).add('Default', () => (
```
hborghols: Would be nice if you could also add `disabled` and `label` to the story

src/components/Slider/test.tsx
```diff
@@ -0,0 +1,100 @@
...
+    it('should not crash when no onChange is provided', () => {
+        const changeMock = jest.fn();
+        const component = mountWithTheme(<Slider value={5} minLimit={1} maxLimit={10} />);
+
+        component
+            .find('input')
+            .at(0)
+            .simulate('change', 1)
+            .simulate('blur');
+        expect(changeMock).not.toHaveBeenCalled();
```
hborghols: This doesn't test if it crashes. You can expect the component not to throw, that asserts if an error has occured.

src/components/Slider/test.tsx
```diff
@@ -0,0 +1,100 @@
...
+                return <div />;
+            },
+        );
+
+        mountWithTheme(<Slider value={5} minLimit={1} onChange={onChangeMock} maxLimit={10} />);
+
+        expect(onChangeMock).toHaveBeenCalledWith(3);
+    });
+
+    it('should be able to change value with slider 2', () => {
```
hborghols: I don't think this testcase is necessary and in any case you're not asserting if it is `able to change value with slider`.

src/components/Slider/test.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+        component
+            .find('input')
+            .at(0)
+            .simulate('change', 1)
+            .simulate('blur');
+        expect(changeMock).not.toHaveBeenCalled();
+    });
+
+    it('should be able to change value with slider', () => {
```
hborghols: Maybe you could also still move the slider

src/components/Slider/test.tsx
```diff
@@ -0,0 +1,100 @@
...
+        component
+            .find('input')
+            .first()
+            .simulate('change', { target: { value: '90' } })
+            .simulate('blur');
+
+        expect(component.find(InputSlider).props().value).toBe(15);
+    });
+
+    it('should not crash when no onChange is provided 2', () => {
```
hborghols: Here it is! I think this is sufficient to test this case. Imo you can remove the first test.

src/components/Slider/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+            <StyledWrapper grow={1} focus={inputFocus} disabled={props.disabled ? props.disabled : false}>
+                <InputRange
+                    value={inputValue}
+                    disabled={props.disabled}
+                    onChange={(value: number) => {
+                        setInputFocus(false);
+                        setInputValue(value);
+                        if (props.onChange !== undefined) props.onChange(value);
+                    }}
+                    minValue={Math.floor(props.minLimit)}
```
hborghols: You could memoize this like in `Range`

hborghols:
> Using the arrows in the number input it's possible to go beyond the min/max values

I get that the value isn't passed back btw but it would be nice (IMO) to also prevent the number in the input from going up. Maybe you can check with UX which scenario is desired.

src/components/Slider/style.tsx
nicokoelewijn: Good idea, i'll check 👍

src/components/Slider/index.tsx
nicokoelewijn: It's unlikely that the minValue will change during its life-cycle. But i'll add it to the inputValue :)

src/components/Slider/test.tsx
nicokoelewijn: Actually, this functionality is tested in react-input-range package ([link](https://github.com/davidchin/react-input-range/blob/master/test/input-range/input-range.spec.jsx)) so perhaps its a bit redundant to test this here as well?

src/components/Slider/style.tsx
nicokoelewijn: The verdict is no. It should have its own type

src/components/Slider/test.tsx
hborghols: I suppose so :)

src/components/Slider/story.tsx
```diff
@@ -1,8 +1,15 @@
 import Slider from '.';
 import React from 'react';
 import { storiesOf } from '@storybook/react';
-import { number } from '@storybook/addon-knobs';
+import { number, text, boolean } from '@storybook/addon-knobs';
 
 storiesOf('Slider', module).add('Default', () => (
-    <Slider value={10} minLimit={number('minValue', 5)} maxLimit={number('maxValue', 25)} onChange={() => undefined} />
+    <Slider
+        label={text('aria-label', 'slider')}
```
hborghols: Oh sorry didn't realize this was only used for the aria label 😓. In that case it's not necessary in the story.

src/components/Slider/story.tsx
```diff
@@ -1,8 +1,15 @@
...
+        value={10}
+        minLimit={number('minValue', 5)}
+        maxLimit={number('maxValue', 25)}
+        onChange={() => undefined}
+        disabled={boolean('disabled', false)}
```
hborghols: Nice 👍, however the inputField doesn't get the disabled visual state yet when disabling the slider. Functionally it works well!

LuukHorsmans:
I approve, but don't have sufficient rights to click the checkmark :P

hborghols:
> I approve, but don't have sufficient rights to click the checkmark :P

I gotchu fam

PR #404, opened 2019-07-16
Additional Slider props

nicokoelewijn:
### This PR:

resolves [STRF-160](https://myonlinestore.atlassian.net/browse/STRF-160)

- Component `Slider` now also supports a couple of optional props: `hideInputField`, `onFocus`, `onBlur`

**Checklist** 🛡

### DIFF
--- a/src/components/Slider/test.tsx
+++ b/src/components/Slider/test.tsx
@@ -41,7 +41,11 @@ describe('Slider', () => {
 
     it('should not crash when no onChange is provided', () => {
         const fn = (): void => {
-            mountWithTheme(<Slider value={5} minLimit={2} maxLimit={7} onChange={undefined} />);
+            const component = mountWithTheme(<Slider value={5} minLimit={2} maxLimit={7} onChange={undefined} />);
+            component
+                .find('input')
+                .first()
+                .simulate('change', { target: { value: '2' } });
         };
 
         expect(fn).not.toThrow();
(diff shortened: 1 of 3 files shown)


src/components/Slider/test.tsx
```diff
@@ -47,6 +47,22 @@ describe('Slider', () => {
         expect(fn).not.toThrow();
     });
 
+    it('should not crash when no onFocus is provided', () => {
```
CarloPalinckx: I see you copied some bugs over from the test suite of Range. These three tests (including onChange) can be improved a bit imo:

- You are still passing in callbacks here. Even though the returns are undefined. There is still a callback. In your code you check for props.onBlur to be undefined or not.
- The coverage report shows that you aren't testing focus and blur behaviour yet. Would be awesome if you could test those callbacks too.

Screenshot 2019-07-16 at 11 24 54

src/components/Slider/test.tsx
```diff
@@ -41,12 +41,57 @@ describe('Slider', () => {
...
         };
 
         expect(fn).not.toThrow();
     });
 
+    it('should not crash when no onFocus is provided', () => {
+        const fn = (): void => {
+            mountWithTheme(<Slider value={5} minLimit={2} maxLimit={7} onFocus={undefined} />);
+        };
+
```
CarloPalinckx: don't forget to also simulate the focus/blur behaviour, it is never going to call props.onBlur etc if you are not actually triggering a blur.

PR #408, opened 2019-07-19
Added baseline for docs website

CarloPalinckx:
### This PR:

Instead of using storybook as our presentational docs I suggest we could have a go at creating a simple static website that is rebuild on every deploy. This could also be used to run puppeteer against to get some SSR tests going.

**Checklist** 🛡


hborghols:
Do you want to get rid of storybook completely or keep it around for dev purposes (like in SF3)? Because I do like it for that.

CarloPalinckx:
> Do you want to get rid of storybook completely or keep it around for dev purposes (like in SF3)? Because I do like it for that.

Haven't really thought about that. This is just visitor-facing. This would also allow us to configure storybook more towards development I guess.

hborghols:
Yeah exactly 😄. I'm definitely in favor of having something nicer than storybook as a "showcase"!

PR #430, opened 2019-08-27
Icon name docs

woutsluiter:
### This PR:

Adds documentation for assets

### DIFF
--- a/src/assets/README.md
+++ b/src/assets/README.md
@@ -1,7 +1,7 @@
 # Assets
 ## Filenaming
 Filenames should be a good description of the icon itself, not its action in an application. 
-For example: an icon of a trash can should be something like `trash.svg`, not `delete.svg`.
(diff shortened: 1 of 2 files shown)


src/assets/README.md
```diff
@@ -0,0 +1,8 @@
+# Assets
+## Filenaming
+Filenames should be a good description of the icon itself, not its action in an application. 
+For example: an icon of a trash can should be something like `trash.svg`, not `delete.svg`.
```
hborghols: why not `trashcan.svg`? 😄

woutsluiter: Hehe true, but our trash can icon file is named `trash.svg` 😄

hborghols: Fair enough, but maybe another example is more suitable in that case. One that fully embraces the desired naming conventions 😛

PR #435, opened 2019-09-04
Added the Carousel component

### LINKED ISSUE #429: Carousel component
Currently there is not really a standardised way to show a carousel in any of our applications. Because this a pattern that can be useful across multiple applications. I suggest we bring it over here. 

I suggest a component with the following api:

```jsx
// controlled example
const App = () => {
    const [currentIndex, setIndex] = useState(0);

    return (
        <Carousel
            loop // wether the carousel returns to the first slider after pressing next on the last slide. (and vice-versa)
            currentIndex={currentIndex} // index of current slide. Can also be undefined if you don't want to control the index yourself.
            onChange={newIndex => {
                setIndex(newIndex);
            }}
            // optional, onChange would probably suffice in most use cases
            onForward={newIndex => {
                setIndex(nextIndex);
            }}
            // optional, onChange would probably suffice in most use cases
            onBackwards={newIndex) => {
                setIndex(nextIndex);
            }}
            // optional
            onClick={clickedIndex => {
                // can be used to highlight image for example
            }}
        >
            // any reactnode could be used in the carousel
            <span>slide 0</span>
            <span>slide 1</span>
            <span>slide 2</span>
        </Carousel>
    );
};
```

```jsx
// uncontrolled example
const App = () => {
    return (
        <Carousel loop>
            <span>slide 0</span>
            <span>slide 1</span>
            <span>slide 2</span>
        </Carousel>
    );
};
```

### Things I'd like feedback on 🎡
The carousel commonly has dots accompanying them. Not sure if we should internalise these are if we should have a separate Dots component. The dots can be pretty complex themselves because they need to be "paginated" (like on instagram) to prevent 100 of dots.

### Additional context 🖼
There are currently 2 designs for this:

Screenshot 2019-08-26 at 15 29 42
Screenshot 2019-08-26 at 15 29 32

Maybe @LuukHorsmans could chime in on the hover, click behaviour etc.

CarloPalinckx:
### This PR:

resolves #429

**Breaking changes** 🔥
- None

- The new component `Carousel` has been added.

- None

**Checklist** 🛡

### DIFF
--- a/src/components/Carousel/index.tsx
+++ b/src/components/Carousel/index.tsx
@@ -4,6 +4,8 @@ import IconButton from '../IconButton';
 import chevronRight from '../../assets/icons/chevron-right.svg';
 import chevronLeft from '../../assets/icons/chevron-left.svg';
 
+const SLIDE_TIMEOUT = 300;
+
 const OuterWrapper = styled.div<{ ratio: [number, number] }>`
     position: relative;
     overflow: hidden;
@@ -19,7 +21,7 @@ const Slide = styled.div<{ active: boolean; x: string }>`
     z-index: ${({ active }) => (active ? '1' : '')};
     transform: translateX(${({ x }) => x});
     transition-timing-function: ease;
-    transition: transform 0.3s;
+    transition: transform ${SLIDE_TIMEOUT}ms;
 `;
 
 const SlideButton = styled(IconButton)<{ direction: 'prev' | 'next' }>`
@@ -46,9 +48,13 @@ type PropsType = {
      */
     slide?: number;
     /**
-     * Allows you to add a callback. It'll receive the slide that should be shown **next**.
+     * This callback will be fired at the **before** the animation. It receives the destination of the slide first, and the origin as second argument.
      */
-    onChange?(newSlide: number): void;
+    onChange?(destination: number, origin: number): void;
+    /**
+     * This callback fires after the animation is complete, allowing you to do something in the time between the onChange and onAfterChange.
+     */
+    onAfterChange?(): void;
 };
 
 const Carousel: FC<PropsType> = props => {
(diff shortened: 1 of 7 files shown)


CarloPalinckx:
@RianneSchaekens can I get a design review for this? :)

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     * This component has 2 modes, controlled and uncontrolled.
+     * You can manage the state yourself by passing in the active slide
+     * and adding an onChange callback
+     */
+    slide?: number;
+    /**
+     * Allows you to add a callback. It'll receive the slide that should be shown **next**.
+     */
+    onChange?(newSlide: number): void;
```
Dreamdealer: Since this callback is called **before** the next slide is animated in maybe it's better to explicitly name this callback like that. I think this is a nicer name since the exact moment when a callback is fired could be important in some cases.

```suggestion
    onBeforeChange?(newSlide: number): void;
```

src/components/Carousel/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    return (
+        <OuterWrapper ratio={ratio} data-testid={props['data-testid']}>
+            {slides.map((child, index) => {
+                return (
+                    <Slide x={`${(index - slide) * 100}%`} key={index} active={slide === index}>
+                        {child}
+                    </Slide>
+                );
+            })}
+            <SlideButton
```
Dreamdealer: Shouldn't these buttons be conditional and only be shown when there is more than 1 slide?

CarloPalinckx: Ah right! Didn't consider the single slide state

src/components/Carousel/index.tsx
CarloPalinckx: This is a bit of a tricky one, because you also want to use this callback to control the state of the current slide if you need to control the carousel:

```jsx
<Carousel 
    slide={slide}
    onChange={(newSlide) => {
        setSlide(newSlide);
    }} 
/>
```
I think in this scenario I think `onChange` is the most obvious naming (since we use that on a plethora of other components).

I definitely think animation based callbacks like onBeforeChange/onAfterChange can be useful. I can imagine a case like:

```jsx
<Carousel 
    slide={slide}
    onChange={(newSlide, oldSlide) => {
        setSlide(newSlide);
    }} 
    onBeforeChange={(newSlide, oldSlide) => setAnimating(true)}
    onAfterChange={(newSlide, oldSlide) => setAnimating(false)}
/>
<Button title={`Buy product ${products[slide].title}`} disabled={isAnimating} />
```

I suggest we both implement these, what do you think @Dreamdealer ?

src/components/Carousel/index.tsx
CarloPalinckx: Ah right! Didn't consider the single slide state

PR #444, opened 2019-09-30
Close persistent toast on primary action

hborghols:
### This PR:

The persistent prop was added to `Toast` to be able to use the toast in scenarios where they are forced to interact with it. In the way it was added it was not possible to close a remotely triggered persistent toast. This PR executes the close action on the primary button action to close the toast after a "successful" interaction has taken place. This is in line with our currently (single) usecase for the persistent `Toast`.

- Persistent `Toast` closes on primary button action

**Checklist** 🛡
- [ ] Appropriate tests have been added for my functionality (check if not applicable).

### DIFF
--- a/src/components/Toast/index.tsx
+++ b/src/components/Toast/index.tsx
@@ -32,7 +32,14 @@ type ButtonVariant = 'primary' | 'destructive' | 'warning' | 'secondary' | 'plai
 
 class Toast extends Component<PropsType> {
     private action = (): void => {
-        if (this.props.onClick !== undefined) this.props.onClick();
+        if (this.props.onClick !== undefined) {
+            this.props.onClick();
+
+            // When a toast is persistent, executing the primary action closes it instead of one of the other closing behaviors
+            if (this.props.persistent) {
+                this.closeAction();
+            }
+        }
     };
 
     private closeAction = (): void => {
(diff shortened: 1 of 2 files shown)


CarloPalinckx:
This is such a specific usecase to build in the component :/ I also don't really get the persistent prop tbh, the Toast has an "autoDismiss" prop that can be set to false. There must be a better way to accomplish this. I think:

>  In the way it was added it was not possible to close a remotely triggered persistent toast

We should solve instead of baking it in.

hborghols:
What would you propose? The only solution I see is copying the toast/toaster functionality to it's own specific component. This is possible of course but also seems like a waste of time imo.

CarloPalinckx:
Perhaps you could look into an api similar to something like setTimeout where there are a few possible solutions:

Solution 1:
```
const id = window.toaster.notify({...props});
window.toaster.close(id);
```

Solution 2: 
```
const close = window.toaster.notify({...props});
close()
```

CarloPalinckx:
You would have to adjust this default behaviour a bithttps://github.com/MyOnlineStore/bricks/blob/master/src/components/Toaster/index.tsx#L86. You could check if autoDismiss is defined or not, when undefined you could maintain the severity behaviour and if defined you could use the specified prop.

hborghols:
I agree, opening this PR was a bit premature

PR #450, opened 2019-10-14
Refactored textfield story + added component support to affix props

Dreamdealer:
### This PR:

**Breaking changes** 🔥
- None

- The `prefix` and `suffix` props on `TextField` now also take ReactNodes instead of only strings
- The TextField component story refactored from class to FC and cleaner code with less duplication

**Checklist** 🛡

Result:
image

Left is a Checkbox, right a clickable IconButton.

### DIFF
--- a/src/components/TextField/test.tsx
+++ b/src/components/TextField/test.tsx
@@ -72,14 +72,15 @@ describe('TextField', () => {
         const component = mountWithTheme(
             <TextField
                 data-testid="foo"
-                prefix={<Box>Foo</Box>}
-                suffix={<Box>Bar</Box>}
+                prefix={<Box data-testid="prefix">Foo</Box>}
+                suffix={<Box data-testid="suffix">Bar</Box>}
                 value=""
                 name="foo"
                 onChange={jest.fn()}
             />,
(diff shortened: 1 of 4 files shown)


src/components/TextField/test.tsx
```diff
@@ -66,4 +67,19 @@ describe('TextField', () => {
...
+                data-testid="foo"
+                prefix={<Box>Foo</Box>}
+                suffix={<Box>Bar</Box>}
+                value=""
+                name="foo"
+                onChange={jest.fn()}
+            />,
+        );
+
+        expect(component.find('[data-testid="foo"]').hostNodes().length).toBe(1);
```
CarloPalinckx: I don't think this is the expect you are looking for .
isntwhatyourelookingfor

I suggest adding a data-testid to the pref and suffix, that way you can expect that these 2 are rendered properly.

Dreamdealer: ah yes, good thinking... i figured that if the two components were provided and they wouldn't render, the entire textfield wouldn't be rendered either. That's just stupid. I'll refactor this.

PR #451, opened 2019-10-14
Add clear TextField button

Dreamdealer:
### This PR:

resolves #BIER-1250

- Added an optional `showClearButton` prop
- This prop shows a button that returns an empty string in the onChange callback when clicked

**Checklist** 🛡

**WARNING: this PR is based on SCS-1065.**

### DIFF
--- a/src/components/TextField/index.tsx
+++ b/src/components/TextField/index.tsx
@@ -110,6 +110,7 @@ class TextField extends Component<PropsType, StateType> {
                                         this.forceFocus();
                                     }}
                                     style={{ cursor: 'pointer' }}
+                                    data-testid={`${this.props['data-testid']}-clear-button`}
                                 >
                                     <Icon icon={CloseSmallIcon} color="#A6AAB3" size="small" />
                                 </div>
--- a/src/components/TextField/index.tsx
+++ b/src/components/TextField/index.tsx
@@ -9,6 +9,7 @@ import Icon from '../Icon';
 import lockedIcon from '../../assets/icons/locked.svg';
 import questionCircle from '../../assets/icons/question-circle.svg';
 import dangerCircle from '../../assets/icons/danger-circle.svg';
+import { CloseSmallIcon } from '../../assets';
 
 type PropsType = {
     value: string;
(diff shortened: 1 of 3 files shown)


src/components/TextField/index.tsx
```diff
@@ -23,9 +24,10 @@ type PropsType = {
     suffix?: string | ReactNode;
     disabled?: boolean;
     placeholder?: string;
+    showClearButton?: boolean;
     'data-testid'?: string;
     extractRef?(ref: HTMLInputElement): void;
-    onChange(value: string, event: ChangeEvent<HTMLInputElement>): void;
+    onChange(value: string, event?: ChangeEvent<HTMLInputElement>): void;
```
CarloPalinckx: Making this event optional would be a breaking change :/
Imagine a user doing this before a release

```
<TextField value="A" onChange={(value, event) => {
  console.log(event.target.value); // this throw type errors and also runtime errors, because event is possibly undefined.
}}/>
```
I think there is an elegant solution to this though.

Maybe instead of adding `showClearButton` you could instead add an `onClear` callback which does not contain an event in the callback `onClear(): void;` as a user you could still easily achieve the same effect by calling your state-setter with an empty string. You could even also show something like a toast confirming that the input was cleared. You could then conditionally render the button if the callback is set or not.

src/components/TextField/index.tsx
```diff
@@ -100,7 +102,19 @@ class TextField extends Component<PropsType, StateType> {
...
+                            {this.props.showClearButton && !this.props.disabled && this.props.value !== '' && (
+                                <div
+                                    onClick={() => {
+                                        this.props.onChange('');
+                                        this.forceFocus();
+                                    }}
+                                    style={{ cursor: 'pointer' }}
+                                    data-testid={`${this.props['data-testid']}-clear-button`}
+                                >
+                                    <Icon icon={CloseSmallIcon} color="#A6AAB3" size="small" />
```
CarloPalinckx: maybe you could move this color into an uppercased constant since it's being used twice here.

src/components/TextField/index.tsx
Dreamdealer: Yeah, I struggled with this one too, but didn't think the effects through of making this optional. Thanks for the explanation. I'll make a fix for this!

PR #457, opened 2019-10-31
Renamed loading prop on button base element

woutsluiter:
### This PR:

removes the `loading` prop on the button base DOM element. The passing of this prop to the dom element is unnecessary and causes warnings in the latest `styled-components` version:

```
Warning: Received `false` for a non-boolean attribute `loading`.
If you want to write it to the DOM, pass a string instead: loading="false" or loading={value.toString()}.
```

**Breaking changes** 🔥
- none

- none

- Removed `loading` prop to button base DOM element

**Checklist** 🛡

### DIFF
--- a/src/components/Button/base/index.tsx
+++ b/src/components/Button/base/index.tsx
@@ -29,7 +29,6 @@ const ButtonBase: FunctionComponent<PropsType> = (props): JSX.Element => {
                 title={props.title}
                 className={props.className}
                 id={props.id}
-                loading={props.loading}
                 data-testid={props['data-testid']}
                 onMouseEnter={props.onMouseEnter}
                 onMouseLeave={props.onMouseLeave}
@@ -41,14 +40,12 @@ const ButtonBase: FunctionComponent<PropsType> = (props): JSX.Element => {
 
     return (
         <StyledButton
-            {...props}
             title={props.title}
             className={props.className}
             onClick={clickAction}
             type="button"
             disabled={props.disabled}
             id={props.id}
-            loading={props.loading}
             data-testid={props['data-testid']}
         >
             {props.children}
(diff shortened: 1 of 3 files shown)


woutsluiter:
hmm wait I think the styling now no longer works 🧐. But the non-HTMLprops should filtered out somehow. Back to the drawing board.

CarloPalinckx:
Works on your deployment 🤷‍♂

yarn.lock
```diff
@@ -26,6 +26,13 @@
   dependencies:
```
CarloPalinckx: Could you revert these changes?

woutsluiter:
> Works on your deployment 🤷‍♂

loading specific styling from the base component doesn't, like the cursor should be `default` instead of `pointer`

yarn.lock
woutsluiter: this is an update of `styled-component` and all its dependencies

src/components/Button/base/index.tsx
```diff
@@ -41,15 +41,14 @@ const ButtonBase: FunctionComponent<PropsType> = (props): JSX.Element => {
...
-            {...props}
             title={props.title}
             className={props.className}
             onClick={clickAction}
             type="button"
             disabled={props.disabled}
             id={props.id}
-            loading={props.loading}
             data-testid={props['data-testid']}
+            data-loading={props.loading ? 'true' : 'false'}
```
CarloPalinckx: this still gets set on the html element, you are better of just naming the prop "isLoading" to avoid these issues

CarloPalinckx: > this still gets set on the html element, you are better of just naming the prop "isLoading" to avoid these issues

PR #491, opened 2020-02-03
Refactored CurrencyField, props now properly apply

### LINKED ISSUE #364: TextField does not always update the rendered value when value prop changes
### Bricks version 📦
0.12.0

### Describe the bug 🔍
TextField does not always update the rendered value when value prop changes. 

### To Reproduce 💥
Set a timeout or interval that updates state.
Observe a stale TextField (Currency)

### Expected behavior ✨
Value should update when props change without an onChange being called.

### Screenshots 📸

Apr-24-2019 10-07-17

CarloPalinckx:
### This PR:

resolves #364 

**Breaking changes** 🔥
- none

- none

- When changing the value prop of the TextField.Currency, the change is now properly applied on the underlying TextField and input

**Checklist** 🛡


Dreamdealer:
- When disable negative numbers prop is true, the cursor jumps to the end of the value when trying to insert a - symbol somewhere.
- `allowDecimals` does not work. I can still enter decimals when this is unchecked 
-  `maximumFractionDigits` and `minimumFractionDigits` do not work. There are always two decimals. Also: do you want to be able to change this with currency? isn't it always 2 decimals?
- Placeholder is never displayed. When clearing the field it reverts back to 0. Shouldn't it be possible to be able to clear the field? IE, when the product editor would use this field people, can accidentally  set the price of a product to zero.
- Prefix en suffix fields are ignored. Doesn't matter what you fill in, there's always a $ prefix. Changing the currency changes the displayed prefix.
- When I change the locale, it reverts back to `en-US` after half a second.
- When entering `5.10` and refocussing the field it changes the value to `5.1`. I think this is confusing for users. Can you check with UX if this is okay?

CarloPalinckx:
> allowDecimals does not work. I can still enter decimals when this is unchecked

This prop only applies to the number field, the story is structured a bit weirdly...

CarloPalinckx:
> Prefix en suffix fields are ignored. Doesn't matter what you fill in, there's always a $ prefix. Changing the currency changes the displayed prefix.

This one is a bit tricky imo. You can never reliably set either a prefix or suffix on this field because you wont know what side the currency symbol is alligned. I'll discuss with UX what they recommend we do here 👍

CarloPalinckx:
>  Placeholder is never displayed. When clearing the field it reverts back to 0. Shouldn't it be possible to be able to clear the field? IE, when the product editor would use this field people, can accidentally  set the price of a product to zero.

This was already the case before the refactor. Perhaps you could leave a seperate issue for that?

PR #516, opened 2020-03-05
Move GlobalStyles from _document to _app

hborghols:
### This PR:

~It seemed like using `createGlobalStylesheet` from styled-components in the `_document` of the DS site didn't always play nicely with Next, causing some pages to be exported with missing fonts. Since this global style was only used to include font-faces I replaced it with Next's recommended (https://nextjs.org/docs/basic-features/built-in-css-support) way to include stylesheets and removed the custom `_document` altogether.~

~As a bonus this also seems to resolve the weird "flashing" issue in chrome on dev when switching between components in a preview pane.~

See @CarloPalinckx comment below. `GlobalStyles` has been moved from `_document` to `_app`

- Replaced custom `_document.tsx` and `createGlobalStylesheet` with `styles.css`.

**Checklist** 🛡


CarloPalinckx:
I don't think this is the proper solution for doing this. Pretty sure this removes styled-components SSR, which might end up giving you "mismatch in className" I think your problem here is that you are including the server stylesheet along with the global styles in the _document component. Styled components does SSR like this:

- Next generates SSR react tree
- S-C pulls all styles from that tree (including global styles) 
- S-C renders the stylesheets in your <head>
- Styles get rendered.

I think your previous approach added a step to it:

- Next generates SSR react tree
- S-C pulls all styles from that tree (including global styles) 
- S-C renders the stylesheets in your <head>
- Styles get rendered.
- GlobalStyles do not get pulled out from the tree but are rendered when hydrating
- Styles get rerendered again.
- Funky buzzinezz

You can probably solve the original problem by moving your global styles from _document to _app. Which would mean that you global styles will get properly extracted for the SSR tree, and S-C doesn't need to reapply them.

CarloPalinckx:
Can confirm, moving styles to app solves fonts 👍 doesn't solve the flickering tho...

hborghols:
> Can confirm, moving styles to app solves fonts 👍 doesn't solve the flickering tho...

Too bad but it is what it is then regarding the flickering. Will add your suggestion for the other problem.

packages/components/docs/pages/_app.tsx
```diff
@@ -11,17 +12,37 @@ declare module 'react' {
...
+      font-family: Melbourne;
+      src: url('/static/fonts/melbourne.ttf');
+      font-weight: 400;
+  }
+  @font-face {
+      font-family: Melbourne;
+      src: url('/static/fonts/melbourne-light.ttf');
+      font-weight: 300;
+  }
+`;
```
CarloPalinckx: nit:
```suggestion
`;

```

PR #518, opened 2020-03-10
feat: implement 3.0 Button design

hborghols:
### This PR:

**Breaking changes** 🔥
- `MultiButton` has been removed
- Button now supports an `info` variant
- New disabled states
- Button ThemeType now requires `info` variant, `border` instead of `textDecoration` in the plain button theme object, `disabled` theme split in `primary`, `secondary` and `plain` variants.

- Freshly styled buttons 🌈 

- There are no longer overrides for `Button` in `Contrast`

**Checklist** 🛡


LuukHorsmans:
I found three things that should be fixed: 

1. The border-radius of the disabled button is incorrect. See the following screenshot
image

2. Please add cursor types to disabled ('not-allowed') and loading ('progress')

3. Lline-heights and paddings for all buttons (including buttons with icon+text) in the design are
    line-height: 24px;
    padding: 6px 24px;
Both line-height and padding seem to be off in this PR and they seem to be different for buttons with icon and buttons without icon. They should be the same.

packages/components/src/components/Button/index.tsx
```diff
@@ -116,25 +130,46 @@ const StyledButton = styled(Base)<PropsType>`
...
-                background: ${theme.Button.disabled.backgroundColor}
-                    repeating-linear-gradient(
-                        -45deg,
-                        ${theme.Button.disabled.stripingColor},
-                        ${theme.Button.disabled.stripingColor} 10px,
-                        transparent 10px,
-                        transparent 20px
-                    );
+                ${
+                    variant === 'plain'
```
woutsluiter: I think it would be clearer to use a switch (or else if statements) and group these properties per variant. Nested ternaries are less clear IMO.

packages/components/src/components/Button/index.tsx
```diff
@@ -116,25 +130,46 @@ const StyledButton = styled(Base)<PropsType>`
...
-                color: ${theme.Button.disabled.color};
-                border-color: transparent;
+                background: ${
+                    variant === 'plain'
+                        ? theme.Button.disabled.plain.backgroundColor
+                        : variant === 'secondary'
+                        ? theme.Button.disabled.secondary.backgroundColor
+                        : theme.Button.disabled.primary.backgroundColor
+                };
+
```
woutsluiter: nit: newline

packages/components/src/components/Button/index.tsx
```diff
@@ -116,25 +130,46 @@ const StyledButton = styled(Base)<PropsType>`
...
+                        : variant === 'secondary'
+                        ? theme.Button.disabled.secondary.color
+                        : theme.Button.disabled.primary.color
+                };
+
+                ${variant === 'plain' ? `border: ${theme.Button.disabled.plain.border};` : ''}
+
                 transform: none;
-                box-shadow: none;
+
```
woutsluiter: nit: newline

PR #555, opened 2020-05-19
feat: Added Card component

woutsluiter:
### This PR:
Adds the new Card component 🃏. I changed and added a few props though:
- ~`enabled`~ > `checked`, because it would be weird with the following prop:
- +`disabled`, to disable the toggle
- +`onChange`, to have a callback for when the toggle is changed
- +`onClick`, to have a callback when the Card is opened/closed

**Breaking changes** 🔥
- none

- Added Card component
- Added `data-testid` attributes to Toggle

- Added `.gypi` files to .gitignore that seem to be generated when local Xcode and node-gyp work propperly

**Checklist** 🛡


woutsluiter:
I see your point. I find it difficult here to strike a balance between forcing design rules and separating concerns. The padding for label, toggle, description and content for instance are all different and easy to miss. Just like the rule that a description should always fold with the content. I think these are rules you don't want to be bothered with every time, but you do want to be consistent.

Another option is to approach this more as a composition that lives in the Canopy backoffice package as a component. Then I would split this up in Card and Accordion.

CarloPalinckx:
> Then I would split this up in Card and Accordion and mature this composition in Canopy until we know all the use cases well enough to promote it to Bricks (or not).

That sounds great! That way we could see if it is usable beyond the current use-case this was built for.

woutsluiter:
Gonna split this up in `Card` and `Accordion` for now

PR #570, opened 2020-06-26
chore: Added PR Title action

woutsluiter:
### This PR:

Adds a PR Title action in favor of the GitHub pr title app

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -13,4 +13,4 @@ jobs:
     steps:
       - uses: amannn/action-semantic-pull-request@v1.2.0
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-title.yml
```diff
@@ -0,0 +1,16 @@
...
+      - edited
+      - synchronize
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v1.2.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
CarloPalinckx, line 16: The token has a different name for us, I think its `MOS_EU_GITHUB_TOKEN`

woutsluiter: Seems to work, but I'll change it 👍

PR #576, opened 2020-07-16
feat(Select): Refactor Select into more split up components

### LINKED ISSUE #575: SelectModal positioning and content flow issues
### Bricks version 📦
3.3.2

### Describe the bug 🔍
The positioning of a `SelectModal` is not calculated correctly when using custom rendering. It uses the height of the selected option, but should use the height of the search field instead. See screenshots.

Then there is another problem: the height of the selected option is taken out of the content flow when opened. This is because the selected option and the search field are rendered as one component (`SearchInput`), one replacing the other depending on state. Here lies the root of the problem: these two should not be one component at all. The selected option (or empty option) should always be rendered in the content to prevent content flow jumps. The search input has a direct relation to the list of options and thus should be rendered in the same place (or portal) as the SelectModal.

### To Reproduce 💥
1. Add some long content after the `Select` custom rendering story
2. Open the `Select`

### Expected behavior ✨
A select dropdown rendered as a whole.

### Screenshots 📸
image

### Proposed solution 💡
I want to seperate concerns: split up the `SearchInput` into a `SearchInput` and `SelectedOption`. Furthermore I want to split up the `SelectModal` into `SelectModal` and `SelectList`. You would get something like this:

```javascript
<SelectWrapper>
    {/* Only takes care of rendering a selected or empty option */}
    <SelectedOption />
    {/* Only takes care of positioning */}
    <SelectModal>
        {/* Only takes care of filtering */}
        <SelectInput />
        <SelectList />
    </SelectModal>
</SelectWrapper>
```

woutsluiter:
### This PR:

resolves #575 

- Splits up `SelectInput` into `SelectSearch` and `SelectedOption`
- Splits up `SelectModal` into `SelectModal` and `SelectList`

**Checklist** 🛡


packages/components/src/components/Select/SelectSearch.tsx
```diff
@@ -0,0 +1,129 @@
...
+    flex-grow: 1;
+    margin: ${INNER_OFFSET - 1}px;
+    padding: 5px 11px;
+    transition: border-color 150ms;
+    box-sizing: border-box;
+    background: ${({ theme }) => theme.Select.common.background};
+    border-radius: ${({ theme }) => theme.Select.common.borderRadius};
+
+    ${({ theme, focus, open, disabled }) => {
+        if (focus && !open && !disabled) {
```
CarloPalinckx, line 58: You can use early returns here do reduce the complexity of the if statements here

packages/components/src/components/Select/SelectSearch.tsx
```diff
@@ -0,0 +1,129 @@
...
+                            data-testid={props['data-testid'] ? `${props['data-testid']}-input-field` : undefined}
+                            onChange={(event: ChangeEvent<HTMLInputElement>): void => {
+                                event.stopPropagation();
+                                setFilter(event.target.value);
+                            }}
+                        />
+                    </SelectSearchInner>
+                    <StyledCaret inner>
+                        <Icon
+                            title={'close'}
```
CarloPalinckx, line 117: Nit:
```suggestion
                            title="close"
```

packages/components/src/components/Select/SelectedOption.tsx
```diff
@@ -0,0 +1,118 @@
...
+    align-items: stretch;
+    transition: border-color 150ms, box-shadow 150ms, background 150ms;
+    box-sizing: border-box;
+    width: 100%;
+    border-radius: ${({ theme }) => theme.Select.common.borderRadius};
+    line-height: 1.6; // results in 24px with 15px fontSize
+    user-select: none;
+
+    ${({ theme, focus, open, disabled }) => {
+        if ((focus || open) && !disabled) {
```
CarloPalinckx, line 28: Could also look into early returns here

packages/components/src/components/Select/test.tsx
```diff
@@ -399,7 +399,7 @@ describe('Select', () => {
 
         expect(
             component
-                .find('[data-testid="foo-window-empty"]')
+                .find('[data-testid="foo-list-empty"]')
```
CarloPalinckx, line 402: you could also use findByTestId here to get rid of the hostNodes and css selector here

packages/components/src/components/Select/test.tsx
```diff
@@ -76,14 +76,14 @@ describe('Select', () => {
...
-        expect(component.find('[data-testid="select-input"]')).toHaveStyleRule(
+        expect(component.find('[data-testid="select-field"]')).toHaveStyleRule(
             'border',
             `solid 1px ${mosTheme.Select.select.focus.borderColor}`,
         );
 
         component.find(Select).simulate('blur');
 
-        expect(component.find('[data-testid="select-input"]')).toHaveStyleRule(
+        expect(component.find('[data-testid="select-field"]')).toHaveStyleRule(
```
CarloPalinckx, line 86: you could also use findByTestId here to get rid of the hostNodes and css selector here

packages/components/src/components/Select/test.tsx
```diff
@@ -76,14 +76,14 @@ describe('Select', () => {
 
         component.find(Select).simulate('focus');
 
-        expect(component.find('[data-testid="select-input"]')).toHaveStyleRule(
+        expect(component.find('[data-testid="select-field"]')).toHaveStyleRule(
```
CarloPalinckx, line 79: you could also use findByTestId here to get rid of the hostNodes and css selector here

packages/components/src/components/Select/SelectSearch.tsx
woutsluiter: This was not really changed actually, but sure, I'll be a good boy scout 👍

packages/components/src/components/Select/test.tsx
woutsluiter: That's not standard right? I would need https://www.npmjs.com/package/enzyme-selectors

packages/components/src/components/Select/test.tsx
woutsluiter: That's not standard Enzyme I see. That's https://www.npmjs.com/package/enzyme-selectors right?

packages/components/src/components/Select/SelectSearch.tsx
woutsluiter: I removed the statements here completely since they no longer are necessary

packages/components/src/components/Select/SelectSearch.tsx
```diff
@@ -34,6 +35,10 @@ const SelectSearchContainer = styled.div`
             color: ${({ theme }) => theme.Select.select.focus.placeholderColor};
             opacity: 1;
         }
+
+        &::ms-clear {
```
CarloPalinckx, line 39: good als microsoft selectors 🙏

packages/components/src/components/Select/index.tsx
```diff
@@ -250,42 +250,45 @@ const Select = <GenericOptionType extends OptionBaseType>(props: PropsType<Gener
...
-                                              option.value === targeted ? '-targeted' : ''
-                                          }`
-                                        : undefined
-                                }
-                            >
-                                {props.renderOption?.(option, optionState)}
-                            </SelectOption>
-                        );
-                    }) || props.children}
+                    <SelectSearch
```
CarloPalinckx, line 265: This is great! Maybe we could start looking into exposing these to the user side in the next major version, that way you we don't have to prop drill all the props for SelectSearch and SelectList. Would also give the user more freedom on the composition of things.

packages/components/src/components/Select/test.tsx
CarloPalinckx: yep! That package adds it to enzyme's prototype and types it as well so you should be able to use it here.

CarloPalinckx: Oh wait, nevermind. I thought I have already installed it in bricks, but I haven't. Feel free to add it if you want (Or don't :) whichever you prefer)

woutsluiter: I'll add it to Bricks then :)

packages/components/src/components/Select/SelectSearch.tsx
CarloPalinckx: Aah this is a renamed file, right. Thanks!

woutsluiter:
I made a separate issue for that one 👉 #577

packages/components/src/components/Select/index.tsx
woutsluiter: Yeah could be a possibility. Though I can't see a use case right now other than having the option of hiding the search field. But there might be more easy options there.

packages/components/src/components/Select/test.tsx
```diff
@@ -60,7 +60,7 @@ describe('Select', () => {
             key: ' ',
         });
 
-        expect(component.find('[data-testid="select-window-open"]').hostNodes()).toHaveLength(1);
+        expect(component.findByTestId('select-window-open').hostNodes()).toHaveLength(1);
```
CarloPalinckx, line 63: You can remove all the hostNodes here since findByTestId already returns just the hostNodes :)

woutsluiter: Ugh geez 🙄

PR #629, opened 2021-05-06
chore: Added Toolbar to FileInput Component

Dreamdealer:
Please ignore the actual toolbar. This is going to be replaced with the new separate component @CarloPalinckx created just now.

### DIFF
--- a/packages/components/src/components/FileInput/index.tsx
+++ b/packages/components/src/components/FileInput/index.tsx
@@ -3,6 +3,7 @@ import SeverityType from '../../types/SeverityType';
 import Box from '../Box';
 import Text from '../Text';
 import Icon from '../Icon';
+import Toolbar from '../Toolbar';
 import IconButton from '../IconButton';
 import InlineNotification from '../InlineNotification';
 import {
@@ -11,7 +12,7 @@ import {
     GearIcon,
     TrashIcon,
 } from '@myonlinestore/bricks-assets';
-import { StyledWrapper, StyledFileInput, StyledPreviewImage, StyledToolbar } from './style';
+import { StyledWrapper, StyledFileInput, StyledPreviewImage } from './style';
 import { ThemeContext } from 'styled-components';
 
 export type InputSeverityType = 'error';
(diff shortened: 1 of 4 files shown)


CarloPalinckx:
Would be nice if we could "inverse control" here by accepting a <Toolbar> component through a `props.toolbar`

CarloPalinckx (changes requested):
Would be nice if we could "inverse control" here by accepting a <Toolbar> component through a `props.toolbar`

packages/components/src/components/FileInput/index.tsx
```diff
@@ -73,14 +68,19 @@ const FileInput: FC<PropsType> = props => {
...
-                    />
+                    <>
+                        <Box grow={1} shrink={1} justifyContent="center" alignItems="center">
+                            <StyledPreviewImage
+                                src={image.source}
+                                alt={image.alt}
+                                style={{ maxHeight: `calc(${props.maxHeight} - 24px` }}
+                            />
+                        </Box>
+                        <Box style={{ zIndex: 2 }}>{props.toolbar}</Box>
```
CarloPalinckx, line 82: This box can be optional I guess
```
{props.toolbar && <Box style={{ zIndex: 2 }}>{props.toolbar}</Box>}
```

packages/components/src/components/FileInput/index.tsx
```diff
@@ -28,23 +29,17 @@ export type PropsType = {
     feedback?: FeedbackType;
     placeholder: ReactNode;
     dropPlaceholder: ReactNode;
+    toolbar: ReactNode;
```
CarloPalinckx, line 32: I'd make this optional for whyen you want to use it without toolbar

CarloPalinckx (changes requested):

packages/components/src/themes/MosTheme/MosTheme.theme.ts
```diff
@@ -33,6 +33,7 @@ const theme: ThemeType = {
             fontSize: fontSize.base,
             fontFamily: bodyFont,
             iconColor: colors.green500,
+            toolbarBackground: colors.grey100,
```
CarloPalinckx, line 36: This doesn't seem needed anymore

CarloPalinckx (changes requested):

PR #643, opened 2021-05-31
feat: Added new ColorField component

Dreamdealer:
### This PR:

- Added new ColorField component

**Checklist** 🛡

<details>

  - Use `fix: ...` for patches

</details>

### DIFF
--- a/packages/components/src/components/ColorField/index.tsx
+++ b/packages/components/src/components/ColorField/index.tsx
@@ -1,0 +1,33 @@
+/// <reference path="../../_declarations/global.d.ts" />
+import TextField, { PropsType as TextFieldPropsType } from '../TextField';
+import React, { FC, useState, useRef } from 'react';
+
+type OmittedKeys = 'onChange' | 'value' | 'prefix';
+
+type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>> & {
+    value: string;
+    onChange(value: string): void;
+};
+
+const ColorField: FC<PropsType> = props => {
+    const [displayValue, setDisplayValue] = useState(props.value);
+    const inputRef = useRef<HTMLInputElement>();
+
+    return (
+        <TextField
+            {...props}
+            extractRef={ref => {
+                inputRef.current = ref;
+                if (ref && props.extractRef !== undefined) props.extractRef(ref);
+            }}
+            value={displayValue}
+            prefix="#"
+            onChange={value => {
+                setDisplayValue(value);
+                props.onChange(value);
+            }}
+        />
+    );
+};
+
+export default ColorField;
--- a/packages/components/src/components/ColorField/index.tsx
+++ b/packages/components/src/components/ColorField/index.tsx
@@ -6,10 +6,7 @@ import Box from '../Box';
 
 type OmittedKeys = 'prefix';
 
-type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>> & {
-    value: string;
-    onChange(value: string): void;
-};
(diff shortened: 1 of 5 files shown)


packages/components/src/components/ColorField/index.tsx
```diff
@@ -0,0 +1,38 @@
+/// <reference path="../../_declarations/global.d.ts" />
+import TextField, { PropsType as TextFieldPropsType } from '../TextField';
+import React, { FC, useState, useRef } from 'react';
+import ColorDrop from '../ColorDrop';
+import Box from '../Box';
+
+type OmittedKeys = 'onChange' | 'value' | 'prefix';
```
CarloPalinckx, line 7: You only have to Omit `prefix` here since you don't have to override the types of onChange en value since they are the same for TextField as ColorField

packages/components/src/components/ColorField/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+    onChange(value: string): void;
+};
+
+const ColorField: FC<PropsType> = props => {
+    const [displayValue, setDisplayValue] = useState(props.value);
+    const inputRef = useRef<HTMLInputElement>();
+
+    return (
+        <Box alignItems="center">
+            <ColorDrop color={`#${displayValue}`} style={{ marginRight: '12px' }} />
```
CarloPalinckx, line 20: Prefer a Box over inline styles

packages/components/src/components/ColorField/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+            <TextField
+                {...props}
+                extractRef={ref => {
+                    inputRef.current = ref;
+                    if (ref && props.extractRef !== undefined) props.extractRef(ref);
```
CarloPalinckx, line 25: Nit: Could this be multiline? 👼

packages/components/src/components/ColorField/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+type OmittedKeys = 'onChange' | 'value' | 'prefix';
+
+type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>> & {
+    value: string;
+    onChange(value: string): void;
+};
+
+const ColorField: FC<PropsType> = props => {
+    const [displayValue, setDisplayValue] = useState(props.value);
```
CarloPalinckx, line 15: I don't think you need state here? Since this is a controlled component you can use props.value instead right?

CarloPalinckx (changes requested):

packages/components/src/components/ColorField/index.tsx
Dreamdealer: Sure. This came from the copy/paste from CurrencyField. I'll fix it in that file too.

packages/components/src/components/ColorField/index.tsx
```diff
@@ -0,0 +1,48 @@
+/// <reference path="../../_declarations/global.d.ts" />
+import TextField, { PropsType as TextFieldPropsType } from '../TextField';
+import React, { FC, useRef } from 'react';
+import ColorDrop from '../ColorDrop';
+import Box from '../Box';
+
+type OmittedKeys = 'prefix';
+
+type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>> & {
```
CarloPalinckx, line 9: You can remove the 
```
& {
    value: string;
};
```
here since we are Picking the props from `TextFieldPropsType` which also have these 2 defined with strings

CarloPalinckx (changes requested):

CarloPalinckx (approved):

PR #647, opened 2021-06-03
chore: Added initialValue and ResetButton to ColorField

Dreamdealer:
- Added initialValue prop and ResetButton

**Checklist** 🛡

<details>

  - Use `fix: ...` for patches

</details>

### DIFF
--- a/packages/components/src/components/ColorField/index.tsx
+++ b/packages/components/src/components/ColorField/index.tsx
@@ -1,14 +1,17 @@
 /// <reference path="../../_declarations/global.d.ts" />
 import TextField, { PropsType as TextFieldPropsType } from '../TextField';
-import React, { FC, useRef } from 'react';
+import React, { FC, useRef, useState } from 'react';
 import ColorDrop from '../ColorDrop';
 import Box from '../Box';
 
 type OmittedKeys = 'prefix';
 
-type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>>;
+type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>> & {
+    initialValue: string;
+};
 
 const ColorField: FC<PropsType> = props => {
+    const [showResetButton, setShowResetButton] = useState(props.value !== props.initialValue);
     const inputRef = useRef<HTMLInputElement>();
 
     // Since the ColorField uses a prefix for the # character, it needs to strip the # from the value when
--- a/packages/components/src/components/ColorField/index.tsx
+++ b/packages/components/src/components/ColorField/index.tsx
@@ -1,6 +1,6 @@
 /// <reference path="../../_declarations/global.d.ts" />
 import TextField, { PropsType as TextFieldPropsType } from '../TextField';
-import React, { FC, useRef, useState } from 'react';
+import React, { FC, useRef } from 'react';
 import ColorDrop from '../ColorDrop';
 import Box from '../Box';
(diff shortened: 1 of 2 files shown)


packages/components/src/components/ColorField/index.tsx
```diff
@@ -1,14 +1,21 @@
...
 
-type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>>;
+type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>> & {
+    initialValue: string;
+    resetButtonTitle: string;
+    onChange(value: string): void;
+};
 
 const ColorField: FC<PropsType> = props => {
+    const [showResetButton, setShowResetButton] = useState(props.value !== props.initialValue);
```
CarloPalinckx, line 18: I think you can avoid using state here since that condition is all you need to show/hide the button

CarloPalinckx (changes requested):

packages/components/src/components/ColorField/index.tsx
Dreamdealer: This is a genuine facepalm moment. Thanks! Hahaha

packages/components/src/components/ColorField/index.tsx
```diff
@@ -33,11 +39,22 @@ const ColorField: FC<PropsType> = props => {
...
                     value={stripHashtag(props.value)}
                     prefix="#"
-                    onChange={(value, event) => {
-                        props.onChange(`#${value}`, event);
+                    onChange={value => {
+                        props.onChange(`#${value}`);
                     }}
                 />
             </Box>
+            <Box width="38px">
```
CarloPalinckx, line 47: Small nit: The IconButton seems to be 36px wide instead of 38px

packages/components/src/components/ColorField/index.tsx
```diff
@@ -3,10 +3,16 @@ import TextField, { PropsType as TextFieldPropsType } from '../TextField';
...
+import { UndoIcon } from '@myonlinestore/bricks-assets';
+import IconButton from '../IconButton';
 
 type OmittedKeys = 'prefix';
 
-type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>>;
+type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, OmittedKeys>> & {
+    initialValue: string;
+    resetButtonTitle: string;
+    onChange(value: string): void;
```
CarloPalinckx, line 14: It's back 😱

CarloPalinckx (changes requested):

packages/components/src/components/ColorField/index.tsx
Dreamdealer: That's on purpose. The original onChange prop also requires an onChange event as second param. The reset button doesn't have this event, but the TextFields props require it. This is how i fixed it. This is also how it's done in the CurrencyField...

packages/components/src/components/ColorField/index.tsx
Dreamdealer: It is, but with 36 it moves. 38 results in a fixed size. Probably some bordersizing issue? Didn't spend a lot of time on this.

Also: I asked Rianne about this fixed size. Might get reverted in the future.

packages/components/src/components/ColorField/index.tsx
CarloPalinckx: Ah fair enough, good call 👍

CarloPalinckx (approved):

PR #652, opened 2021-06-07
chore: Added tests to ColorInput component

Dreamdealer:
### This PR: 

Adds tests.

### DIFF
--- a/packages/components/src/components/ColorField/index.tsx
+++ b/packages/components/src/components/ColorField/index.tsx
@@ -18,10 +18,9 @@ type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, Omit
 };
 
 export enum TestIds {
-    colorInputContainer = 'color-input-container',
-    colorInputField = 'color-input-field',
+    container = 'container',
     colorDrop = 'color-drop',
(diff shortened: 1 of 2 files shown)


packages/components/src/components/ColorField/index.tsx
```diff
@@ -17,6 +17,13 @@ type PropsType = Pick<TextFieldPropsType, Exclude<keyof TextFieldPropsType, Omit
     onChange(value: string): void;
 };
 
+export enum TestIds {
+    colorInputContainer = 'color-input-container',
```
CarloPalinckx, line 21: These cannot be static like they are in canopy since you have to be able to pass in a data-testid through props. Would be nice if the id passed in through props is that one that currently corresponds with `colorInputField` so that you can directly simulate on that in tests. The other ones can be a composite of the prop plus a suffix (`foo-reset`) for instance

CarloPalinckx (changes requested):

packages/components/src/components/ColorField/index.tsx
Dreamdealer: Yes. I make that.

CarloPalinckx (approved):

PR #667, opened 2021-07-12
fix: Fixed z-indexing issues and added a close callback to window

Dreamdealer:
### This PR:

Fixed z-indexing issues and added a close callback to window. Also added box props to ColorPicker component.

**Checklist** 🛡

<details>

  - Use `fix: ...` for patches

</details>

### DIFF
--- a/packages/components/src/components/TextField/index.tsx
+++ b/packages/components/src/components/TextField/index.tsx
@@ -99,7 +99,7 @@ const TextField: FC<PropsType> = props => {
                         <StyledAffix>{props.prefix}</StyledAffix>
                     </StyledAffixWrapper>
                 )}
-                <Box position="relative" width="100%">
+                <Box position={props.onClear && props.value !== '' ? 'relative' : undefined} width="100%">
                     <StyledInput
                         data-testid={props['data-testid']}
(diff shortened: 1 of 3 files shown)


packages/components/src/components/TextField/index.tsx
```diff
@@ -99,7 +99,7 @@ const TextField: FC<PropsType> = props => {
                         <StyledAffix>{props.prefix}</StyledAffix>
                     </StyledAffixWrapper>
                 )}
-                <Box position="relative" width="100%">
+                <Box position={props.onClear && props.value !== '' ? 'relative' : undefined} width="100%">
```
CarloPalinckx, line 102: Why was this added? 🤔

CarloPalinckx (changes requested):

packages/components/src/components/TextField/index.tsx
Dreamdealer: The position relative is only used for the pos:abs that the reset/clear button uses. And since the reset button is not used in the colorpicker and that is what causes the z-index issues, i choose to only apply the position relative if it is needed.

CarloPalinckx (approved):

PR #683, opened 2021-09-29
feat!: Fileinput now keeps preview around on re-render

CarloPalinckx:
### This PR:

The FileInput currently does not support being re-rendered with a file, the file preview is removed once the FileInput is re-rendered. This PR solves this problem by lifting the ref of the image out of the component with the `value` prop. A ref is passed in that can have its current value read in the onChange callback.

**Breaking changes** 🔥
- The FileInput now requires a value prop to be set up in order to function
- Moved some props around to prevent nesting in the props

- None

- Added a test and a story to cover the preview behaviour

**Checklist** 🛡

<details>

  - Use `fix: ...` for patches

</details>


CarloPalinckx:
For the merge message:
> BREAKING CHANGE: FileInput now requires a ref to be passed into the value prop

woutsluiter (changes requested):
Looking good! I noticed a small flash of the original preview prop image when you toggle the image. Is that because it's a fixed value in the story? 

One thing though: the `Disabled` story breaks with this change.
image

CarloPalinckx:
The preview prop is often used as a fixed prop that you rarely adjust so this shouldn't be flashing. My guess would be due to the async behaviour of the FileReader. I'll look into it 👍

woutsluiter (changes requested):
Now when I change the file in the `WithToggle` story it disappears the first time 🤔

CarloPalinckx:
> Now when I change the file in the WithToggle story it disappears the first time 🤔

Aah because the ref changes after picking a file the props get recomputes. I memoized it which fixed it ✅

woutsluiter (approved):

PR #697, opened 2021-11-04
fix(Icon): Prevent calling setState after unmount

woutsluiter:
### This PR:

image

**Breaking changes** 🔥
- none

- none

- Cancel promise and prevent calling `setIcon` when `Icon` component is unmounted.

**Checklist** 🛡

<details>

  - Use `fix: ...` for patches

</details>


packages/components/src/components/Icon/index.tsx
```diff
@@ -22,14 +25,19 @@ const Icon: FC<PropsType> = (props): JSX.Element => {
                     }
                 })
                 .then(output => {
-                    if (typeof output === 'string') {
+                    if (typeof output === 'string' && isSubscribed) {
```
CarloPalinckx, line 28: Instead of preventing the setState you could maybe prevent the promise from resolving at all by aborting the request on unmount with something like [this](https://davidwalsh.name/cancel-fetch). Added benefit of this is that when navigating away while the icons are loaded you don't have to wait for all the icons to resolve before navigating away.

CarloPalinckx (changes requested):

packages/components/src/components/Icon/index.tsx
woutsluiter: Even better ✨

packages/components/src/components/Icon/index.tsx
```diff
@@ -12,10 +12,12 @@ type PropsType = typeof box.props & {
 
 const Icon: FC<PropsType> = (props): JSX.Element => {
     const [icon, setIcon] = useState<string>('');
+    const controller = new AbortController();
```
CarloPalinckx, line 15: Can you either move this into the useEffect or use a useRef for these 2? I think they get overwritten by additional renders otherwise

CarloPalinckx (changes requested):

packages/components/src/components/Icon/index.tsx
woutsluiter: hmm good one 🤔

CarloPalinckx (approved):

PR #752, opened 2022-04-14
feat: Added hoverColor to Link component

CarloPalinckx:
### This PR:

- Hovercolor now supports a `hoverColor` prop

- Refactored stories to CSF

**Checklist** 🛡

<details>

  - Use `fix: ...` for patches

</details>


woutsluiter:
Could you explain the use-case for the prop? Since bricks will be MOS brand only I would argue this should be a pattern and therefore another variant. Or perhaps the default should be changed?

woutsluiter:
> I'd suggest that perhaps we could looks into deprecating all the link variants based on severity and instead implements those with design tokens like: `<Link hoverColor={colors.blue800} />`.
> 
> Does that sound like something that would work?

The fixed variant hover colors are a bit weird indeed, but I don't think `hoverColor` is the right solution to the issues you mentioned. The thing I would want Bricks to do is: 
- make it easy to build stuff
- while guaranteeing the MOS brand styling. 

With a `hoverColor` prop I would have to figure out which color is the right one each time I use it. It makes it more error prone.

Deprecating the variants might not be a bad idea. But I think we would have to discuss this with UX to alter the current styling slightly. The way I see it there are two types of use cases of the Link component:
1. In a paragraph with a theme (hover)color
2. In a context (like Notification or Footer) inheriting the color

If we keep the hover color the same when in context and only adjust the text-decoration you could refactor it with:
- default styling with themed (hover)color
- inherit behavior with an `inherit` prop or something

Perhaps we can discuss this next FE meeting.

CarloPalinckx:
Closing this for now, another approach is in the works

PR #758, opened 2022-05-24
feat: Set focus on modal on show

### LINKED ISSUE #423: Improve modal focus behaviour
When opening a modal, it gets a very low tab priority. This means that when you open it and want to tab through it, you have to go through the entire page first.

The best solution would be to trigger a focus on the modal container when the open prop is changed.

Alternatively we could set the focus to the button group instantly. Which doesn't cover all cases because there could be focusable content in the modal as well. This would require a prop to set the focus behaviour.

### Additional context 🖼

Watch me having to tab through all the rows in this example ⬇️ 
Aug-12-2019 10-06-17

KuiseRapper:
### This PR:

resolves #423 

**Breaking changes** 🔥
- None

- Modal component now sets tabindex 0 and focus on modalwrapper.

- None

**Checklist** 🛡

<details>

  - Use `fix: ...` for patches

</details>


packages/components/src/components/Modal/index.tsx
```diff
@@ -87,8 +87,17 @@ const Modal: FC<PropsType> = props => {
         };
     });
 
+    useEffect(() => {
+        styledModalWrapperRef.current?.focus();
```
CarloPalinckx, line 91: This also focusses the wrapper when `props.show` goes from `true` -> `false`. You want to wrap this in an if statement to avoid the modal getting focussed when closed

CarloPalinckx (changes requested):

packages/components/src/components/Modal/test.tsx
```diff
@@ -242,4 +242,39 @@ describe('Modal', () => {
         expect(component.findByTestId('modal-close-button').length).toBe(1);
         expect(component.findByTestId('modal-buttons-container').length).toBe(1);
     });
+
+    it('Should have TabIndex 0 on show', () => {
```
CarloPalinckx, line 246: Last nit: "Should" should be lowercased 😅

packages/components/src/components/Modal/test.tsx
```diff
@@ -242,4 +242,39 @@ describe('Modal', () => {
...
+                    <Button key="foo" variant="primary" title="foo" />,
+                    <Button key="bar" variant="primary" title="bar" />,
+                ]}
+            />,
+        );
+
+        expect(component.find(StyledModalWrapper).prop('tabIndex')).toBe(0);
+    });
+
+    it('Should have undefined (no) TabIndex on show', () => {
```
CarloPalinckx, line 262: Last nit: "Should" should be lowercased 😅

CarloPalinckx (changes requested):

CarloPalinckx (approved):

PR #843, opened 2022-09-20
chore: CASH-3283 - Rewrite Slider tests

hborghols:
### This PR:

🎚️ 

Testing-library does not actually render the slider component from `react-input-range` so we can't test the props of that component. I removed those tests for now.


packages/components/src/components/Slider/test.tsx
```diff
@@ -1,105 +1,94 @@
...
         expect(focusMock).not.toHaveBeenCalled();
 
-        component.find('input').first().simulate('focus');
+        const inputField = screen.getByTestId('slider-input');
 
-        expect(focusMock).toHaveBeenCalled();
-    });
+        fireEvent.focus(inputField);
 
-    it('should floor and ceil the minLimit and maxLimit respectively', () => {
```
woutsluiter, line 91: Isn't this something you'd still want to guarantee?

woutsluiter (approved):

PR #854, opened 2022-09-21
fix: Only set text-align on Text when prop is defined

hborghols:
### This PR:

- Doesn't set `text-align` when the prop is undefined so that the Text component inherits the parent's alignment.

**Checklist** 🛡

<details>

  - Use `fix: ...` for patches

</details>


packages/components/src/components/Text/index.tsx
```diff
@@ -133,7 +133,9 @@ const Text = styled.p<PropsType>`
 
         return '24px';
     }};
-    text-align: ${({ textAlign }): string => (textAlign ? textAlign : '')};
```
woutsluiter, line 136: 💡 You could also just return the value. I believe the property is omitted when undefined in SC

woutsluiter (approved):

packages/components/src/components/Text/index.tsx
hborghols: I kind of prefer making it explicit rather than relying on SC "magic" tbh

